PR #32, opened 2026-07-01
fix(engines/web): install host metrics bridge before loading d2 wasm

boomsi:
The d2 web loader skipped `installHostMetricsBridge()`, unlike the mermaid and plantuml web loaders. Without it, `globalThis.supramark.measureText` was undefined when d2 rendered alone, so the wasm fell back to the `size*0.6` per-char heuristic — producing layouts that differed from the mermaid→d2 path, where mermaid had already installed the bridge.

Install the bridge at the top of `loadWebD2Render` so both paths use real canvas measurement.

Add `__tests__/web-d2-bridge.test.ts` to pin the invariant that the d2 web loader must install the bridge before its wasm runs.

RN is unaffected: it uses a separate bridge backed by native FFI / Skia.

Behavior change: d2 SVG output now uses real text metrics in all cases, so diagrams that previously rendered with the fallback heuristic (notably container title placement) will shift to match upstream d2.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -117,4 +117,14 @@ ignore = [
     # url -> plantuml-little (duplicate idna in tree). Low real risk for a
     # markdown/diagram toolchain; revisit when url/plantuml dedupe idna.
     "RUSTSEC-2024-0421",
+
+    # ttf-parser is unmaintained (RUSTSEC-2026-0192); the author has
+    # stopped maintenance and future work moved to the skrifa / fontikos
+    # ecosystem. Core parsing dep of font-metrics (TtfParserMetrics) and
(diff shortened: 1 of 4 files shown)


kookyleo (approved):
Reviewed: correct and minimal. d2 was the only web loader missing installHostMetricsBridge (plantuml/mermaid already install it before their wasm import); the call is idempotent (guarded, won't clobber an existing bridge) and placed before the d2 import. RN is unaffected (native FFI path, DOM-guarded). The three nits (host-bridge.ts accidental reindent, the test asserting call-presence rather than ordering, and the unrelated deny.toml advisory) are non-blocking follow-ups.

PR #52, opened 2026-07-03
fix(plantuml): render while-loop condition, loop-back, and branch labels (#38)

### LINKED ISSUE #38: PlantUML 活动图，渲染错误
问题描述：
PlantUML 活动图，渲染错误

测试代码：
```text
@startuml
start
:Open queue;
while (queue empty?) is (no)
  :Handle message;
  :Acknowledge message;
endwhile (yes)
:Close queue;
stop
@enduml
```
实际效果：

Image

官方渲染效果：

Image

mybee:
Split out from #44 (per-issue PRs requested).

## Problem

The `while (cond) is (label)` / `endwhile (label)` construct was only partially implemented in the activity layout:

- The condition diamond used the bare `render_diamond` helper, which drew only the polygon — **condition text was dropped**.
- `build_edges` connected while → body → endwhile → next as a straight vertical line — **no loop-back arrow**.
- The `is` / `endwhile` branch labels were never placed.

Result: a while loop rendered as a straight chain of nodes with empty diamonds and no visible cycle.

## Fix

**layout/activity.rs**: introduce a `WhileFrame` stack + `while_loopbacks` record. `While` pushes a frame (condition text only); `EndWhile` pops it and records the pair. New Pass 3d stamps the `is` label on the while→body edge, the `endwhile` label on the endwhile→next edge, and appends a 4-point snake loop-back edge (endwhile right → right side → while right). New `build_while_loopback_edge` for the geometry; its `extra_right` is folded into `loopback_extra_right` so the viewport reserves room.

**render/svg_activity.rs**: `render_diamond` now centres `node.text` inside the diamond (11pt), gated on `!text.is_empty()` so the empty `repeat`-start diamond is unaffected.

## Verification

Rendered via the wasm package. Before: empty diamonds, straight chain, no cycle. After: condition text `queue empty?`, `no`/`yes` labels, visible loop-back snake — matching the official PlantUML server output.

Fixes #38


kookyleo (approved):
Blocker resolved (render_diamond now splits multi-line text), fmt clean, INDEX panic-landmine removed. Added the while-loop reference-SVG golden (7d603085) — the byte-exact suite gates the render now and passes on CI. Good to merge.

PR #56, opened 2026-07-03
fix(d2): container labels overlap top border under host metrics (#27)

### LINKED ISSUE #27: D2 容器/分组，渲染文字server 贴着/压着容器边框
问题描述：
D2 容器/分组，渲染文字server 贴着/压着容器边框
测试代码：
```text
server
# Declares a shape inside of another shape
server.process

# Can declare the container and child in same line
im a parent.im a child

# Since connections can also declare keys, this works too
apartment.Bedroom.Bathroom -> office.Spare Room.Bathroom: Portal
```

实际效果：
Image

官方渲染效果：
Image

mybee:
## Problem

D2 container labels (`OUTSIDE_TOP_CENTER`) rendered touching/overlapping the container's top border under the wasm/host-canvas metrics backend, while the native path was correct (issue #27).

For:
```
server
server.process
im a parent.im a child
apartment.Bedroom.Bathroom -> office.Spare Room.Bathroom: Portal
```

the "server" label baseline sat at `y=64` with the rect top at `y=70` — a 6px gap that left the label descenders overlapping the border.

## Root cause

The renderer places the label baseline at `label_tl.y + font_size` (`svg_render/mod.rs`), which assumes `ascent ≈ font_size`. `label_height` was computed as `m.ascent + m.descent`, but under host-canvas metrics `m.ascent` is the **tight glyph bounding-box ascent** (often < font_size for lowercase / x-height-only labels like "server"), so `label_height` was too small, `label_tl.y` sat too low, and the label descended into the box border.

Native (`D2GoEmulationMetrics`) was unaffected because its `ascent ≈ font_size` by construction — it matches the locked e2e goldens.

## Fix

In `crates/d2-little/src/textmeasure/d2_host_metrics.rs`, compute `single_h` as `size_f + m.descent` instead of `m.ascent + m.descent`. This matches the native convention so `label_height` is consistent with the render formula, restoring the 5px breathing gap above the box. Native path untouched.

## Verification

Rendered via the wasm package:

| label | before (gap) | after (gap) |
|---|---|---|
| server | ~1px (overlap) | ~5px |
| apartment | ~1px (overlap) | ~5px |
| office | ~0px (overlap) | ~5px |

Native reference-SVG / e2e goldens are unaffected (the fix is wasm-path only).

Fixes #27

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,6 +219,27 @@ jobs:
       - name: cargo test -p plantuml-little (lib unit tests)
         run: cargo test -p plantuml-little --lib
 
+  # d2-little's Rust unit tests (text-measurement / Go-emulation ruler /
+  # layout) had no CI job — the "Test & Build" job only exercises the RN
+  # JS wrapper and the wasm build, never `cargo test -p d2-little`. That
+  # left both the native d2_go_emulation unit tests and any host-path
+  # helpers (e.g. `host_label_height::single_line_height`) without
+  # regression coverage. d2-little is self-contained (no graphviz /
+  # fonttools native deps), so a plain `cargo test --lib` is enough; the
+  # heavier byte-exact e2e / cross-validation suites can be a follow-up.
+  d2-rust-tests:
+    name: d2-little Rust tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: cargo test -p d2-little (lib unit tests)
+        run: cargo test -p d2-little --lib
+
   # plantuml-little byte-exact reference-SVG suite (issue #46, "Full" tier).
   # The committed goldens in tests/reference/ are byte-exact against the wasm
   # Graphviz build (@kookyleo/graphviz-anywhere-web@0.1.8), so the suite runs
--- a/crates/d2-little/src/textmeasure/d2_host_metrics.rs
+++ b/crates/d2-little/src/textmeasure/d2_host_metrics.rs
@@ -38,7 +38,7 @@ use std::cell::Cell;
 
 use font_metrics_core::{Measured, Metrics, host_callback::HostCallbackMetrics};
 
-use super::{D2Metrics, MarkdownOptions};
+use super::{D2Metrics, MarkdownOptions, host_label_height::single_line_height};
 use crate::fonts::{Font, FontFamily, FontStyle};
 
 /// Adapter that bridges d2's [`D2Metrics`] surface to the host
@@ -158,8 +158,9 @@ impl D2Metrics for D2HostMetrics {
         let single_h = {
             let m = self.inner.measure(lines[0], family, size_f, bold, italic);
             // Use `size_f` (the requested font size) as the ascent proxy
-            // instead of `m.ascent` (see `single_line_height` doc).
-            single_line_height(size_f, m.ascent, m.descent)
+            // instead of `m.ascent` (see `host_label_height::single_line_height`
+            // doc for why this approximates — not matches — the native path).
+            single_line_height(size_f, m.descent)
         };
         let composed_h = single_h + ((n - 1) as f64) * self.line_height_factor.get() * size_f;
         (max_w, composed_h)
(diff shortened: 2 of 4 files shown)


kookyleo (commented):
## Review: MERGE WITH NITS — the fix is directionally correct and well-targeted at the top-border overlap, but the "matches the native convention" rationale is imprecise and the two new unit tests never execute in any CI or local `cargo test` path.

Verified the full claim chain against source (read-only worktree, `Actrium:main` vs the PR head) and cross-checked the CI check set via `gh pr checks`. I did **not** execute the wasm host path or reproduce the visual overlap — see the scope note at the end.

### What's right
- **Baseline-placement premise is TRUE.** The renderer does place the label baseline at `label_tl.y + font_size`: shape labels `svg_render/mod.rs:2447` and `:2474`, connection labels `:1353`, arrowhead labels `:1449`. The `ascent ≈ font_size` assumption the fix leans on is real.
- **Under-measurement premise is TRUE.** `D2HostMetrics` forwards the host `canvas.measureText` ascent/descent verbatim (`font-metrics/src/host_callback.rs:81-92`; the fallback even hardcodes `ascent: size*0.8`). Those are tight actual-bounding-box values, so for x-height / lowercase labels `ascent < font_size` and the old `ascent + descent` under-measured the single-line height. Raising it is the correct lever for a *top*-border overlap, and doing it via a named helper + doc comment is a clean, minimal change.
- **Multi-line is not broken.** `composed_h = single_h + (n-1)·lhf·size_f`; the change lifts only the first-line base `single_h`, leaving the per-line gap term untouched, so the total grows by one fixed delta rather than scaling per line — structurally consistent with the native ruler's multi-line composition. Width is unaffected (only the `.1` of the tuple changes).

### 🟠 The two new unit tests never run
`d2_host_metrics.rs` is `#![cfg(target_arch = "wasm32")]` (module top). The new `#[cfg(test)] mod tests` uses plain `#[test]`, not `#[wasm_bindgen_test]`. As a result:
- native `cargo test` compiles the module out (`target_arch != wasm32`);
- the CI wasm step is `wasm-pack build` (`scripts/build-wasm.ts`), and `#[cfg(test)]` code isn't compiled during a build; there is no `wasm-pack test` and no `wasm_bindgen_test` anywhere in the crate.

Net: the tests provide zero regression protection for `single_line_height`. Since it's a pure free function, moving it (or just its test module) out from behind the `wasm32` gate would let it run under a normal `cargo test`.

### 🟠 CI blind spot (independent of this PR)
This PR's green checks exercise no d2-little Rust code: the set is License / Lint / JNI / Test & Build (bun + wasm build) / plantuml-little Rust + reference-SVG. There is no d2-little Rust test job, so neither these new tests nor the existing native `d2_go_emulation` unit/e2e tests run in CI. Worth a follow-up to add a `cargo test -p d2-little` job.

### 🟡 "matches the native `D2GoEmulationMetrics` convention" is imprecise
Native single-line height is not `font_size + descent`. Native `measure_precise` returns the ruler's `bounds.h()` (`d2_emulation_metrics.rs:90-92`), and the ruler replaces each glyph's box height with the **face** `ascent + descent` (`d2_go_emulation.rs:382-387`; face metrics at `:216-219`, constant per size, text-independent). So native single-line height = `face_ascent + face_descent`.

For 28px Source Sans Pro: `face_descent = 7.65625` (= ceil(1792·273/1000)/64 — matches the test constant exactly), `face_ascent ≈ 27.5625` (ascender ≈ 984) → native ≈ **35.22**, not 35.65625. The fix uses `font_size` (28) as the ascent proxy (≈ face_ascent, off by ~0.4px) and keeps the host's tight, text-dependent descent instead of the constant face descent. So for a descender-less label (`m.descent ≈ 0`) the fixed height is `28`, still ~7px shorter than native's ~35. The fix corrects the dominant, ascent-side error — exactly the lever for a top-border overlap — but does not actually equal the native box. Suggest softening the doc/test wording from "matches native" to "approximates the native ascent convention".

### 🟡 Second test comment overstates parity
`single_line_height(28.0, 28.0, 7.65625) == 35.65625` only checks the `font_size + descent` arithmetic; it neither runs the native path nor equals the true native height (~35.22). In production the third argument is the host's tight descent, not a face descent, so "matches native path's values" is misleading.

### 🟡 `_ascent` is now a dead parameter
`single_line_height(font_size, _ascent, descent) = font_size + descent` ignores the passed `m.ascent`. Fine as intent documentation, but the signature is slightly misleading.

### ⚠️ Blast radius wider than the title
The change is on `measure_precise` / `measure_text`, so it lifts the reported height of every host-path label (shapes, connections, containers), not just container labels. Height-dependent anchors (`InsideMiddleCenter`, `Outside*`, `Border*` in `label.rs:431-507`) shift accordingly. This moves the host path toward the taller native boxes, so it is very likely an improvement rather than a regression, but the effect isn't container-scoped.

> Verification scope (please note): I verified the entire claim chain against source in a read-only worktree and confirmed the CI check set from `gh pr checks`. I did not execute the tests (they can't run — see the first 🟠), did not build/run the wasm host path, and did not reproduce the visual overlap on a real host canvas. The pixel-level container top-strip math (grid/nested layout) was not traced end-to-end. Whether the fix fully removes the overlap for descender-less container labels (which stay ~7px shorter than native) should be confirmed with a real before/after render.

### To confirm
1. Move `single_line_height` (or its test module) out from behind `#![cfg(target_arch = "wasm32")]` so it runs under `cargo test`?
2. Add a `cargo test -p d2-little` CI job to close the d2-little Rust blind spot?
3. Reword the "matches native convention" doc/test comments, and consider whether the descent should also be face-based for closer native parity?

kookyleo (approved):
Reviewed: host-metrics label height fix is correct (descent term cancels, restoring the 5px gap); CI job addition is purely additive and closes the d2-little Rust blind spot; 851 tests pass, no regression.

PR #59, opened 2026-07-05
fix(engines/mermaid): center node and edge labels (#40)

### LINKED ISSUE #40: Mermaid 流程图，文字不居中，超出边框
问题描述：
Mermaid 流程图，渲染文字文字不居中，超出边框

测试代码：
```text
---
config:
  htmlLabels: false
---
flowchart LR
  subgraph "One"
    a("`The **cat** in the hat`") -- "edge label" --> b{{"`The **dog** in the hog`"}}
  end
```

实际效果：

Image

官方渲染效果：

Image

mybee:
## Summary

Fixes #40 — Mermaid 流程图文字不居中、超出边框。

For `htmlLabels: false` flowcharts (the issue's repro), labels rendered off-center in two ways:

1. **Node text not vertically centered** — sat at the bottom of the box instead of the middle.
2. **Edge-label text not horizontally centered** — sat ~width/2 to the **left** of its edge.

### Root cause

Both artifacts come from markup that `mermaid-little` emits *faithfully matching upstream mermaid*, so the fix belongs in the engine post-processing layer ([`inlineMermaidSvg`](packages/engines/src/mermaid/index.ts)) rather than the wasm crate (which would break byte-exact parity fixtures).

**Node labels** — each `<foreignObject>` is sized to the font's `ascent+descent`, but the inner `<div>` paints at `line-height: 1.5`, i.e. a line box ~1.5× taller. Under the upstream `display: table-cell`, the div grows to the (taller) content height and overflows the foreignObject **downward**, dropping the text to the bottom of the node.

**Edge labels** — the markup is:
```html
<g class="edgeLabel" transform="translate(lx, ly)">          <!-- edge midpoint -->
  <g class="label" transform="translate(-w/2, -h/2)">         <!-- shifts content left by w/2 -->
    <rect class="background" x="-2" width="w+4"/>             <!-- rect centre → edge midpoint ✓ -->
    <text text-anchor="middle">…anchored at x=0…</text>      <!-- text centre → w/2 LEFT of midpoint ✗ -->
```
The `translate(-w/2, …)` re-centers the **background rect** on the edge, but the `text-anchor=middle` text self-centres at local `x=0` and never receives the same shift, so it lands one half-width to the left.

### Fix

- **`centerMermaidHtmlLabels`** — switch the label `<div>` from `display: table-cell` to `display:flex; align-items:center; justify-content:center; height:100%`. The div now pins to the foreignObject box and centers its content; the foreignObject's `overflow="visible"` (already set by `normalizeMermaidHtmlLabels`) keeps taller text from being clipped.
- **`recenterMermaidEdgeLabels`** — drop the `x` shift from the label-group transform (the text already self-centres) and shift the background rect to `x = -width/2` so text and rect share the same centre line.

## Verification

Reproduced the issue's exact input in a headless browser (Chrome) and measured before/after with `getBoundingClientRect`:

| | before | after |
|---|---|---|
| Node text vs node center (vertical) | ~21 px off | **0 px** |
| Edge-label text vs edge midpoint (horizontal) | ~30 px off | **<1 px** |

Unit tests in [`packages/engines/__tests__/mermaid-label-centering.test.ts`](packages/engines/__tests__/mermaid-label-centering.test.ts) pin both transforms by feeding a raw (pre-fix) SVG fragment through `renderMermaidSvg` with the wasm mocked.

`bun run lint`, `bun run build`, and `bun test` (engines) all pass.

### DIFF
--- a/packages/engines/__tests__/mermaid-label-centering.test.ts
+++ b/packages/engines/__tests__/mermaid-label-centering.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, it, mock } from 'bun:test';
+
+// Issue #40: mermaid flowchart labels rendered off-center.
+//   - Node text sat at the bottom of its box instead of the vertical
+//     middle (foreignObject height = ascent+descent, but the inner div
+//     paints at line-height 1.5 and overflows downward under
+//     `display: table-cell`).
+//   - Edge-label text sat ~width/2 to the LEFT of its edge (the label
+//     group's `translate(-w/2, …)` recentred the background rect but the
+//     text-anchor=middle text self-centres at x=0, so it never got the
+//     shift).
+//
+// Both are visual artifacts of markup that faithfully mirrors upstream
+// mermaid; supramark recenters them in `inlineMermaidSvg`. These tests
+// pin the post-processing by feeding a raw (pre-fix) SVG fragment
+// through `renderMermaidSvg` with the wasm mocked.
+
+const RAW_EDGE_LABEL =
+  '<g class="edgeLabel" transform="translate(200, 100)">' +
+  '<g class="label" data-id="L_a_b_0" transform="translate(-42, -7)">' +
+  '<g><rect class="background" style="" x="-2" y="-2" width="88" height="18"></rect>' +
+  '<text y="-10.1" text-anchor="middle" style="">' +
+  '<tspan class="text-outer-tspan row" x="0" y="-0.1em" dy="1.1em" text-anchor="middle">' +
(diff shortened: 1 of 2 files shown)


kookyleo (approved):
Reviewed and verified against real mermaid-little output: ran both post-processors over all 1394 reference goldens — the node-centering rewrite fires on 832 files (9004 labels) with zero mis-fires, the edge recenter is correctly scoped to the htmlLabels:false text+rect markup (fires only on that exact structure), and before/after screenshots confirm labels center with rendering intact. Failure mode on unexpected markup is a silent no-op, never corruption.

PR #68, opened 2026-07-10
fix(plantuml): sequence group frame width via longest-path gaps (fixes #28)

### LINKED ISSUE #28: PlantUML 时序图，渲染文字压着边框
问题描述：
PlantUML 时序图，渲染文字压着边框
测试代码：
```text
@startuml
Alice -> Bob: Authentication Request
alt successful case
  Bob -> Alice: Authentication Accepted
else some kind of failure
  Bob -> Alice: Authentication Failure
  group My own label
    Alice -> Log : Log attack start
    loop 1000 times
      Alice -> Bob: DNS Attack
    end
    Alice -> Log : Log attack end
  end
else Another type of failure
  Bob -> Alice: Please repeat
end
@enduml
```
实际效果：

Image

官方渲染效果：

Image

mybee:
## Problem

PlantUML sequence diagrams rendered group/`alt`/`loop` frames ~13px too wide, so the frame labels sat flush against the border — the symptom reported in #28.

Root cause: when a message arrow spans a **non-adjacent** participant (e.g. `Alice -> Log`, which crosses `Bob`), the puma layout split the message's required width **evenly across every gap** it spanned. Because the `Alice -> Bob` gap was already very wide (from `Authentication Request`), the `Alice -> Log` messages needlessly inflated the `Bob -> Log` gap, pushing `Log` ~13px to the right and widening every enclosing frame.

## Fix

Match the official PlantUML algorithm. A multi-span message is the same **difference constraint** the official teoz `RealLine` solver uses:

```
posC[hi] − posC[lo] >= needed
```

collected per message and solved by **longest-path** when positioning participants (a single forward pass is exact, since every constraint has `lo < i`). Only the final gap absorbs any deficit; earlier gaps keep their base width. Span-1 (adjacent) messages are unchanged.

## Verification — pixel-level, via the official algorithm

The issue's input now renders **pixel-identical** to official PlantUML `1.2026.7beta3` (the SVG attached to the issue):

| element | official | before | after |
|---|---|---|---|
| Log lifeline x | 297.71 | 310.9526 | **297.71** |
| Bob→Log gap | 50.0567 | 63.2993 | **50.0567** |
| alt frame width | 327.9619 | 341.2046 | **327.9619** |

A full element-by-element diff of the rendered SVG (all 20 texts, 6 frame rects, 3 pentagon tabs, 2 separator lines) against the official output matches exactly. The multi-span gap distribution now also matches the official behaviour — a long `A -> C` message puts its full deficit in the last gap (`[33.6, 275.4]`) instead of splitting evenly (`[154.5, 154.4]`).

## Tests added

- `issue28_participant_positions_match_official` — participant centres match official box centres exactly.
- `issue28_alt_frame_width_matches_official` — alt frame width = 327.9619.
- `issue28_frame_width_matches_official_in_svg` — guards the user-visible symptom at the SVG level.
- `multispan_message_puts_deficit_in_last_gap_3p` / `_4p` — the longest-path distribution invariant.

All 263 reference-SVG tests, the full sequence suite (layout + render), and clippy pass with no regressions.

Fixes #28.

### DIFF
--- a/crates/plantuml-little/src/layout/sequence.rs
+++ b/crates/plantuml-little/src/layout/sequence.rs
@@ -950,6 +950,10 @@ pub fn layout_sequence(sd: &SequenceDiagram, skin: &crate::style::SkinParams) ->
     // the constraint solver ensures centerX >= arrowPreferredWidth. Track this
     // so we can apply it when positioning participants.
     let mut min_first_center: f64 = 0.0;
+    // Multi-participant (span > 1) message constraints, collected as Java teoz
+    // difference constraints `posC[hi] - posC[lo] >= needed` and solved by
+    // longest-path in the positioning pass below. See `span_constraints` use.
+    let mut span_constraints: Vec<(usize, usize, f64)> = Vec::new();
     for event in &sd.events {
         match event {
             SeqEvent::AutoNumber { start } => {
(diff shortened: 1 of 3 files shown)


kookyleo (approved):
Reviewed against source: multi-span messages are modeled as difference constraints solved by longest-path in topological order (a single forward pass is exact). span==1 is byte-equivalent to the prior path. Full plantuml-little suite green (port_sequence_layout 19/0, port_sequence_render 7/0, reference 263/0), no golden regressions. Rebased onto current main; CI green.

PR #75, opened 2026-07-13
fix(plantuml): align if/else activity layout with Java algorithm

### LINKED ISSUE #30: PlantUML 活动图，渲染错误
问题描述：
PlantUML 活动图，渲染错误

测试代码：
```text
@startuml
start
if (Graphviz installed?) then (yes)
  :process all\ndiagrams;
else (no)
  :process only __sequence__ and __activity__ diagrams;
endif
stop
@enduml
```
实际效果：

Image

官方渲染效果：

Image

mybee:
## Summary

Fixes #30 — the `if/else/endif` activity diagram rendered oversized because the branch layout diverged from PlantUML's Java algorithm in three ways. Rather than nudging values, this re-implements the official `FtileIfWithDiamonds` positioning and `FtileIfWithLinks` edge order (reverse-engineered from `plantuml-1.2026.2.jar` via CFR), achieving **byte-identical** output to the official jar for the issue case.

The three root causes:

1. **Box width counted raw creole markup.** `estimate_text_size` measured with raw `text_width`, so `__sequence__` / `__activity__` inflated the else-branch box by ~48px. Java's `SheetBlock1.calculateDimension` strips inline markup (`__`, `**`, `//`) before measuring. Now measures via `creole_text_width` (runs the creole parser), adding back trailing-whitespace width that Java's `SheetBlock1` also preserves — this keeps `activity_mono_multi_line` byte-exact.

2. **Missing merge diamond.** Java emits a `FtileDiamond` (24×24, `getShape2`) at the branch merge point when `hasTwoBranches` (both branches have a normal out-point). We never drew it. Added it at `getYdelta1b` (6px) below the longer branch's actual last-node bottom (`FtileGeometry.outY`).

3. **Wrong horizontal placement + edge routing.** Branches were placed at `diamond_edge ± 10` with ad-hoc edges. Java uses `innerMargin = max(widthInner, diamond1.width + 20)` where `widthInner = box_then_w/2 + box_else_w/2 + 20` (`FtileIfNude`), placing branch1 at `cx - innerMargin/2` and branch2 at `cx + innerMargin/2`. Edges follow `ConnectionHorizontalThenVertical` (diamond→branch) and `ConnectionVerticalThenHorizontal` (branch→merge diamond) snake patterns.

## Scope of the "byte-identical" claim

To be precise about what is and isn't covered by the official-algorithm port:

- **Normal two-branch `if/else/endif` merge** (the issue #30 case, and the common case): fully ported from `FtileIfWithDiamonds` + `FtileIfWithLinks`. Output is byte-identical to `plantuml-1.2026.2.jar` (417×214), verified by element-by-element coordinate diff.
- **`goto`/`break` branches** (no merge diamond, `hasTwoBranches` false): the pre-existing empirical edge routing (`merge_mid_y = merge_y + 5`, etc.) is left untouched. It is not on the official algorithm and is not claimed to be. Issue #30 does not exercise this path. Keeping it unchanged is deliberate to avoid scope creep; it can be ported separately.

## Reference fixture

- `tests/fixtures/activity/if_else_endif.puml` — the issue #30 source verbatim.
- `tests/reference/activity/if_else_endif.svg` — generated from `plantuml-1.2026.2.jar` with `skinparam defaultFontName DejaVu Sans` to force DejaVu metrics on macOS (the existing reference corpus was generated on Linux via fontconfig `sans-serif→DejaVu`; the metrics are identical, as confirmed by the existing `while_loop`/`mono_multi_line` goldens still passing byte-exact). The `font-family="DejaVu Sans"` attribute is normalised to `"sans-serif"` to match the committed reference convention, and the randomised `<?plantuml-src?>` hash PI is stripped (both are no-ops under the test canonicaliser).

## Test plan

- [x] New reference test `reference_fixtures_activity_if_else_endif_puml` passes (byte-exact via canonicalisation)
- [x] All 13 activity reference tests pass — no regressions on `goto_label`, `break`, `backward`, `while_loop`, `mono_multi_line*`, `swimlane001`, etc.
- [x] `cargo test -p plantuml-little --lib` — 2697 passed, 0 failed
- [x] **CI bidirectional-drift check passes locally** (`scripts/check_reference_failures.sh`): 264 passed / 0 failed / 7 ignored — failing set matches the (empty) `known_failures.txt` exactly. The new fixture is the 264th passing test; no new failures, no unexpectedly-healed fixtures.
- [x] `cargo fmt -p plantuml-little --check` clean (the standalone crate CI gates on this)
- [x] `cargo clippy -p plantuml-little --tests -- -D warnings` clean (standalone crate CI gates on this too)
- [x] Scope is limited to `layout/activity.rs` if/else paths — no changes to `while`/`switch`/sequence rendering

## Verification not done from this environment

- The wasm (`plantuml-little-web`) path end-to-end in a browser — exercised the equivalent Rust logic natively via `dump_svg` + the reference suite instead. A live demo is running at `http://127.0.0.1:8765/index.html` (local only) rendering the issue case via a wasm build of this branch if visual confirmation is wanted.


kookyleo (approved):
Reviewed. CI is green and local cargo fmt for plantuml-little passed. Local clippy could not run in my workstation because the Graphviz native library is missing; the GitHub CI jobs cover that path and are green. Approving for merge.

PR #77, opened 2026-07-13
fix(plantuml): byte-exact activity partition+fork rendering (#33)

### LINKED ISSUE #33: PlantUML 活动图，渲染错误
问题描述：
PlantUML 活动图，渲染错误

测试代码：
```text
@startuml
start
partition Client {
  :Submit request;
}
partition Server {
  :Validate request;
  fork
    :Persist result;
  fork again
    :Send audit event;
  end fork
}
partition Client {
  :Render response;
}
stop
@enduml
```
实际效果：

Image

官方渲染效果：

Image

mybee:
## Summary

Fixes the issue #33 case: a PlantUML activity diagram with a `fork`/`fork again`/`end fork` nested inside `partition` blocks rendered incorrectly — missing partition frames, the fork not branching into parallel branches, and a black bar instead of the gray rounded sync bar.

Reaches byte-exact parity with the official Java PlantUML **v1.2026.2** SVG output by reproducing the upstream layout algorithms — **not** by loosening test tolerances. The issue33 golden SVG was generated with the official `plantuml-1.2026.2.jar` (DejaVu fonts) and the implementation matches it byte-for-byte modulo the non-deterministic `<?plantuml-src …?>` source-token PI.

## Key algorithm details

- **Partition frame layout** (Java `FtileGroup` + `USymbolFrame`): `URectangle` with folded-corner tab, title text, `ignoreForCompressionOnX`. Width `= max(content_w + 20, title_w + 20)`; height `= content + diffHeightTitle + 20`.
- **Global X-compression** (`CompressionXorYBuilder`): `SlotFinder` collects occupied intervals (2px edges for ignorable shapes, full width for others); `SlotSet` merges overlapping (inclusive bounds), `reverse()` finds gaps, `smaller(5.0)` drops gaps ≤ 10px and shrinks survivors by 5px/side; `CompressionTransform` applies the delta.
- **`SpecialText`** (`net.atmp.SpecialText`): wraps the partition title when `frame_width - title_w >= 25`. Implements `UShapeIgnorableForCompression`; `drawWhenCompressed` draws `UEmpty(1,1)` at the text end — contributing only a 1px occupied interval rather than the full text width. This is the source of the 0.44px drift a naive width calculation produces.
- **Fork bar width**: sum of per-branch widths (`branch_content_w + 28`), where branch content includes partition frames — not `max_w + 28`.
- **Branch x-positioning**: partition frame nodes are centered within their branch alongside flow nodes, using the range `first..next_branch_start` (or `bottom_bar_idx`).
- **Edge emission order**: reproduces Java's `FtileWithConnection.drawU` post-order traversal — the delegated subtree is drawn first, then the tile's own outgoing connection. Collapsed onto the flat node array as a stable sort by `(top_level_tile_rep(to_index), subgroup)` (subgroup 0 = intra-tile, 1 = inter-tile), which puts an inter-tile connection `[A->B]` after B's internal edges — matching the jar's element order for the issue33 Client→Server→Client partition sequence.
- **X-compression right edge**: for `ignoreForCompressionOnX` shapes (fork bars, partition frames) that span shrink slots, capture the original right edge before transforming `x` and transform *that*, instead of `left + original width`. The old formula double-applied the slot between the left and right edges and over-shrank the bar/frame width.
- **Consecutive partitions**: when the previous flow node sits inside a closed partition, the partition frame extends below the inner node by the bottom margin. Use the partition frame bottom (not the inner node bottom) as the effective bottom when spacing the next partition's frame top, so the three stacked partitions in issue33 no longer overlap.

## Test plan

- [x] `tests/fixtures/activity/issue33.puml` (the exact example from the issue) passes at **STAGE2** (canonical-equal — byte-exact modulo the random `plantuml-src` token). Verified by direct diff: the only difference vs. the jar-generated golden is the `<?plantuml-src …?>` PI.
- [x] New byte-exact golden + fixtures for `fork_simple`, `fork_three_branches`, `partition_simple`, `partition_with_fork`, `fork_with_partition`.
- [x] Full reference suite: **269 passed, 0 failed, 7 ignored** — no regressions.

Closes #33.


kookyleo (changes requested):
Requesting changes before merge: 1. cargo fmt for plantuml-little fails on layout/activity.rs and parser/activity.rs. 2. The new fork/partition layout paths use expect on ForkAgain, EndFork, and PartitionEnd; malformed or unbalanced diagrams should not panic. 3. Fork plus swimlane behavior needs an explicit fix or explicit unsupported-path test because lane replay can shift after ForkAgain stops emitting a layout node. 4. Remove the Co-Authored-By: Claude Sonnet 4.6 trailers, or otherwise ensure they do not land in main history.

kookyleo (approved):
## Re-review: cleanup done — approving

`3008e76f` removes exactly the dead fields flagged last round (ForkFrame's `saved_y_cursor`, PartitionFrame's `diff_height_title` / `saved_last_flow`) and fixes the reference-test count comment; after the main merge the count is consistent again (286 == 286 `reference_test!` entries). The remaining `saved_y_cursor` on `IfFrame` is pre-existing and sits under the crate's deliberate `#[allow(dead_code)] // fields match Java structure` pattern — out of scope.

CI 13/13 green (lib, reference-SVG suite, port integration). As stated last round, this is good to merge.

PR #78, opened 2026-07-13
fix(plantuml): sequence "note right of X" occupies own slot, not arrow overlay (#35)

### LINKED ISSUE #35: PlantUML 时序图，有一部分渲染错误
问题描述：
PlantUML 时序图，有一部分渲染错误

测试代码：
```text
@startuml
participant Browser
participant API
database Cache

Browser -> API: GET /profile
API -> Cache: lookup user
Cache --> API: cached profile
note right of API
  cache hit avoids database
end note
API --> Browser: 200 OK
@enduml
```
实际效果：

Image

官方渲染效果：

Image

mybee:
## Summary

Fixes #35.

PlantUML distinguishes two sequence-note kinds; the port conflated them, which is what produced the wrong rendering in the issue:

- **`Note(null)`** — a *message* note (`note right:` with no `of PARTICIPANT`): attaches to the last arrow, overlaps it, and is wrapped in Java's `ArrowAndNoteBox`.
- **`Note(p)`** — a *participant* note (`note right of X`): a standalone `NoteBox` placed at the current `freeY`, which then advances `freeY` by `noteBox.getPreferredHeight()` via `DrawableSetInitializer.prepareNote()`. It gets its **own vertical slot** and does **not** overlap the previous arrow.

Before this fix, `note right of API` was treated as a message note — rendered overlapping the preceding arrow at the wrong y. This PR routes participant notes through the standalone-slot path that mirrors `prepareNote`, and keeps message notes on the existing back-offset overlay path.

### What changed

- **Model** (`SeqEvent::NoteRight` / `NoteLeft`): add an `on_message: bool` flag.
- **Parser**: set `on_message` from whether an explicit `of PARTICIPANT` clause was written — covers both `note right of X: text` and multiline `note right of X` / `…` / `end note`. (`note right:` / `note right` with no `of` ⇒ message note.)
- **Layout**: participant notes (`on_message == false`) now build a standalone `NoteBox` at `freeY + paddingY` (i.e. `y_cursor − arrow_y_point + NOTE_COMPONENT_PADDING_Y`), x-positioned with the activation right/left shift, and advance `y_cursor` by the NoteBox preferred height — matching Java's `prepareNote`. Mirrored for `note left of X`.
- **Two follow-on correctness fixes** gated on the same flag:
  - `has_regular_msg_note` (the ~3px initial overlay-offset) now fires only for message notes; a participant note doesn't overlay the arrow.
  - The message-id counter (`id="msgN"`) advances only for message notes (ArrowAndNoteBox consumes a slot). Participant notes must not, so ids stay contiguous (`msg1..msg4`), matching the reference.

This is algorithm-level parity with official PlantUML, not empirical coordinate tuning.

## Verification

Pixel-exact against the official **PlantUML v1.2026.2** Docker reference (`eclipse-temurin:21` + `fonts-dejavu-core`):

- With rect participants, the note polygon, note text, every arrow, and the lifelines are **byte-identical** to the reference SVG (`viewBox` 343×243; msg ids `msg1..msg4`).
- The only residual is a `0.0001`pt tail-box text-y difference — a Java float-accumulation quirk (sub-sub-pixel, far inside the reference suite's `2.51`pt tolerance).

### Out-of-scope note

The literal issue puml uses `database Cache`. The `database` cylinder shape's bbox undercounts total width/height by ~10px — a **pre-existing, orthogonal** bug present with or without any note (verified by diffing the same diagram with the note removed). It deserves its own issue/PR, so the new reference fixtures use rect participants to test the note fix in isolation.

## Test plan

- [x] Two new reference fixtures — `issue35_note_right_of.puml`, `issue35_note_left_of.puml` — with **Docker-generated goldens** (`tests/reference/sequence/*.svg`). Both pass.
- [x] Parser unit tests asserting `on_message` for single-line, multiline, and `of`-vs-no-`of` cases (including a multiline `note right` with no `of` ⇒ message note).
- [x] Full crate suite green: **2703** lib + **266** reference (264 prior + 2 new) + **182** integration, **0 failures**, no regressions.
- [x] No new clippy warnings (one pre-existing warning in `ensure_create_participant`, untouched and not CI-gated for this crate).


kookyleo (changes requested):
Requesting changes before merge: 1. cargo fmt for plantuml-little fails on layout/sequence.rs and parser/sequence.rs. 2. Explicit note right/left of X after a self-message is still gated by last_message_was_self and can continue through the message-note path; it should occupy its own participant slot, with regression coverage. 3. note over behavior is internally inconsistent between comments, has_regular_msg_note offset handling, and is_note_on_message false; align behavior/docs and add coverage. 4. Remove the Co-Authored-By: Claude Sonnet 4.6 trailer, or otherwise ensure it does not land in main history.

kookyleo (approved):
Re-reviewed: all four prior CHANGES_REQUESTED items resolved. cargo +1.97.0 fmt -p plantuml-little --check clean; note right/left of X after a self-message now takes its own slot (first-statement `if !on_message { continue; }`), verified against issue35 goldens; note over consistency aligned across has_regular_msg_note / is_note_on_message / comments; no Claude trailer. Tests: lib 2705/0, reference_tests 270/0/7-ignored (6 new byte-exact fixtures), port_sequence_layout 19/0, port_sequence_render 7/0. Non-blocking nit: reference_tests.rs comment count is stale.

PR #83, opened 2026-07-16
fix(rn): use container gap for block spacing, drop trailing margin

boomsi:
## Background

Inside a container (e.g. a chat bubble), the last block (paragraph / heading / code block / ...) had extra space below it, making the container's bottom padding larger than the other three sides.

## Root cause

Each block's default style carried its own `marginBottom` for spacing (paragraph=8, h1=12, table=12, codeBlock=8, list=8, listItem=4, mapCard=12, ...), and `renderNode` did **not** strip the margin on the last block — so the final block always left a trailing margin. `MathBlock` / `DiagramNode` additionally hardcoded `marginVertical` / `marginBottom: 8` (not controllable via the `styles` prop), with the same trailing issue.

## Approach

Move vertical block spacing from per-block `marginBottom` to **container `gap`**. gap only applies between siblings, so the final block has no trailing margin. RN ≥ 0.71 supports gap; this repo already uses `gap:6` in `mapCardContent`.

Effect: `[A,B,C] → nothing after C`; `[A] → nothing after A`; spacing between middle blocks is preserved.

## Changes

- **styles.ts**: `root` gets `flexDirection:column + gap:8`; remove the spacing `marginBottom` from paragraph / h1-h6 / codeBlock / list / listItem / diagramPlaceholder / mapCard / table; `list` gets `gap:4`.
- **Supramark.tsx**: `admonitionContainerStyle` switches to `gap:4`; `defItemStyle` drops marginBottom (covered by `list` gap).
- **MathBlock.tsx** / **DiagramNode.tsx**: drop hardcoded `marginVertical` / `marginBottom`.
- Map card **internal** layout (`mapCardHeader` / `mapCardTitle` / `mapCardContent`) unchanged.

Only affects `@supramark/rn`; the web renderer is not touched.

## Visual changes (FYI)

Top-level block spacing is now a uniform 8: paragraphs stay 8→8; headings (was 12/10), tables / map cards (12), code blocks (8) converge to 8. Multiple footnotes 4→8; admonition inner paragraph spacing 8→4.

## Test plan

- [x] Single-line paragraph: bubble's bottom inset matches top / left / right
- [x] Multiple paragraphs: spacing preserved between, none trailing after the last
- [x] Heading + text / code + text / list + text: normal spacing, no trailing on the last block
- [x] Ordered / unordered lists: normal item spacing, no trailing on the last item
- [x] Both dark and light themes render correctly
- [x] `tsc --noEmit` passes

### DIFF
--- a/packages/renderers/rn/__tests__/spacing.test.ts
+++ b/packages/renderers/rn/__tests__/spacing.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it, mock } from 'bun:test';
+
+// styles.ts import react-native 的 StyleSheet;其 JS 入口含 Flow 语法,bun 无法加载。
+// 用 identity mock —— 测试只关心样式对象的 gap 模型,不依赖 StyleSheet.create 运行时。
+mock.module('react-native', () => ({
+  StyleSheet: { create: (styles: unknown) => styles },
+}));
+
+const { defaultStyles } = await import('../src/styles');
+
+/**
+ * block 间距 gap 模型测试。
+ *
+ * 模型:间距跟着容器走 (gap),不跟着 block 走 (marginBottom)。
+ * - root (column + gap:8) 管 top-level block 间距,无 trailing。
+ * - list (gap:4) 管 list_item 间距。
+ * - 承载 block children 的嵌套容器 (footnote 内层 / definition description / generic container)
+ *   各自在渲染处独立加 gap,不复用共享的 listItem / listItemText —— 后者是 row 布局、不带 gap。
+ */
+describe('block spacing gap model', () => {
+  it('root 用 column + gap 统一 top-level block 间距,无 trailing', () => {
+    expect(defaultStyles.root.flexDirection).toBe('column');
+    expect(defaultStyles.root.gap).toBe(8);
+  });
+
+  it('list 用 gap 给 list_item 间距', () => {
+    expect(defaultStyles.list.gap).toBe(4);
+  });
+
+  it('共享 listItem 是 row 布局且不带 gap (不被嵌套容器污染)', () => {
+    expect(defaultStyles.listItem.flexDirection).toBe('row');
+    expect((defaultStyles.listItem as Record<string, unknown>).gap).toBeUndefined();
+  });
+
+  it('共享 listItemText 不带 gap (不被嵌套容器污染)', () => {
+    expect((defaultStyles.listItemText as Record<string, unknown>).gap).toBeUndefined();
+  });
+
+  it('block 默认样式不再携带 marginBottom (间距改由容器 gap 管)', () => {
+    expect((defaultStyles.paragraph as Record<string, unknown>).marginBottom).toBeUndefined();
+    expect((defaultStyles.h1 as Record<string, unknown>).marginBottom).toBeUndefined();
+    expect((defaultStyles.codeBlock as Record<string, unknown>).marginBottom).toBeUndefined();
+  });
+
+  it('标题用 marginTop 表达"上方比下方宽"的分级 (与 root gap 叠加)', () => {
+    // 上方间距 = root gap(8) + marginTop;下方间距 = root gap(8)
+    expect((defaultStyles.h1 as Record<string, unknown>).marginTop).toBe(8); // 上方 16
+    expect((defaultStyles.h2 as Record<string, unknown>).marginTop).toBe(6); // 上方 14
+    expect((defaultStyles.h3 as Record<string, unknown>).marginTop).toBe(4); // 上方 12
+    expect((defaultStyles.h4 as Record<string, unknown>).marginTop).toBe(2); // 上方 10
+    // h5/h6 不加 marginTop,与上方块等距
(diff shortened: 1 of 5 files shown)


kookyleo (changes requested):
## Review: DO NOT MERGE as-is — undisclosed theme change + two nested containers lose all spacing

The gap approach itself is sound and the core claim holds: `Supramark.tsx:217`'s root View is the direct parent of all top-level blocks, so `gap` applies and the last block no longer leaves a trailing margin; peer requirement (RN >= 0.72) exceeds gap's 0.71 floor. Also verified as non-issues: mapCardHeader/mapCardTitle margins are internal card layout and correctly kept; blockquote/hr have no render path in this renderer, so nothing was missed there; no external consumers reference the removed default margins. tsc/eslint clean; the package's 41 existing tests pass (all svgUtils — none cover layout).

Three items need fixing before merge:

### 🟠 `styles.ts:411-413, 421-423` — dark/light `root.backgroundColor` commented out, not mentioned anywhere in the PR
`darkThemeStyles.root.backgroundColor: '#0d1117'` and `lightThemeStyles.root.backgroundColor: '#ffffff'` are commented out — this has no causal relation to the gap fix and looks like debug residue. With `theme="dark"` the component no longer ships its own dark background, so its light text (#e0e0e0/#fff) renders on whatever the host paints — nearly unreadable on a light host. The example app masks this by painting `#0d1117` itself (`examples/react-native/App.tsx:232`). Please revert, or split into its own PR with a rationale.

### 🟠 `Supramark.tsx:533` — multi-paragraph footnotes collapse to 0 spacing
Footnote content renders block children inside `styles.listItemText` (`{flex:1, lineHeight:20}`, no gap). With paragraph `marginBottom` removed, consecutive paragraphs inside one footnote now touch (was 8). Add a gap to that container.

### 🟠 `Supramark.tsx:441` — definition-list descriptions collapse to 0 spacing
`defDescriptionStyle` is `{paddingLeft:16}` with no gap; a description containing multiple blocks (paragraphs / code) now has zero spacing between them. Dropping `defItemStyle`'s marginBottom is fine (outer `styles.list` gap:4 covers item spacing), but nothing covers inside a description.

### 🟡 Nits

- `Supramark.tsx:412` (and the html card content at 341): the generic container fallback reuses `styles.listItem` (row direction, no gap) as a block container — previously weak spacing there is now zero. Pre-existing debt slightly deepened; worth a gap or a TODO while here.
- Headings converge to a uniform 8 (was h1:12/h2:10/h4:6/h5-h6:4) — acknowledged in the PR, but note a uniform gap can never express "more space above a heading than below"; re-introducing per-block margins would be the only way back.

### ⚠️ Caveats

- styles-prop semantics changed silently: a host's `paragraph: {marginBottom: 12}` used to replace the default 8; it now stacks on top of root gap 8 → 20. The new knob is `root.gap`, but nothing documents this.
- No test in the package covers any of this change; final spacing and the theme background need a simulator before/after check.

kookyleo (approved):
## Re-review: all three blocking items fixed correctly — approving

Item-by-item against the previous round:

1. ✅ Theme `root.backgroundColor` fully restored (`styles.ts:416` dark `#0d1117`, `:426` light `#ffffff`), byte-identical to the pre-PR baseline, with the ownership question properly split out to #88 and the smuggling acknowledged in the commit message.
2. ✅ Multi-paragraph footnotes: `Supramark.tsx:544` now wraps content in `gap: 8` — equivalent to the removed paragraph `marginBottom: 8`.
3. ✅ Definition-list descriptions: `defDescriptionStyle` gains `gap: 8` (`Supramark.tsx:452`), applied on both render paths.

Also addressed beyond the asks: the generic container fallback got a proper column `blockContainerStyle` with gap; heading hierarchy restored via `marginTop` 8/6/4/2 (top spacing 16→10 graded, bottom uniform 8 — marginTop produces no trailing space, so the original bug cannot return; verified no spacing `marginBottom` remains outside card-internal layout); the styles-prop stacking semantics are now documented in JSDoc (`Supramark.tsx:78-85`); and 6 new spacing tests land in `__tests__/spacing.test.ts`.

Local verification: rn package 47 tests pass / 0 fail, `tsc --noEmit` and eslint clean.

### 🟡 Follow-ups (non-blocking)

- A document starting with a heading now gets an 8px leading gap at the root top (`styles.ts:88`) — the top-side mirror of the fixed trailing-margin bug, smaller in magnitude (8 vs the old 12). RN has no `:first-child`; stripping `marginTop` at `index === 0` in renderNode would close it.
- The hierarchy moved from "space below heading" to "space above heading" — typographically sounder, but a visual change the description doesn't call out.
- Header text in the html-card/generic-container paths moved off the `listItemText` slot to inline `{lineHeight: 20}` — hosts customizing `listItemText` via the styles prop no longer reach these; inline style objects are also rebuilt per render.
- `spacing.test.ts`'s `mock.module('react-native')` is process-global under `bun test`; fine today, a hazard if a future test file needs the real module.

Final spacing rhythm and bubble insets still deserve one simulator before/after look, but nothing blocking remains.

PR #86, opened 2026-07-20
fix(diagram): defer rendering for open streamed fences

### LINKED ISSUE #85: 流式 Markdown 中未闭合的图表代码块会反复显示 Rendering 和 Diagram error
## 问题描述

当 Markdown 内容以流式方式持续追加时，如果消息中包含 Mermaid、ECharts、Vega/Vega-Lite、PlantUML、DOT/Graphviz 等 fenced diagram block，在结束围栏尚未到达期间，界面可能在以下状态之间反复切换：

1. `Rendering diagram (...)`
2. `Diagram error: ...`
3. 收到新内容后再次显示 `Rendering diagram (...)`
4. 图表内容完整后最终正常显示 SVG

该现象在使用严格 JSON 格式的 ECharts、Vega/Vega-Lite 图表时尤其明显。

## 复现步骤

1. 使用 `@supramark/rn` 渲染一段逐步增长的 Markdown。
2. 先传入图表开始围栏和部分图表内容，例如：

~~~markdown
```echarts
{
  "xAxis":
~~~

3. 在结束围栏尚未传入时，继续按字符或数据块更新 `markdown`。
4. 观察图表区域状态。
5. 最后传入完整图表内容和结束围栏。

## 实际结果

- 图表内容尚未完整时已经进入图表渲染状态。
- 当前片段无法被图表引擎解析时显示 `Diagram error`。
- 下一次 Markdown 更新后错误消失，再次进入 `Rendering diagram`。
- 流式更新速度和图表引擎返回速度不同时，可能只看到持续 Rendering，也可能看到 Rendering/Error 交替闪烁。
- 完整图表内容到达后最终可以正常显示。

## 预期结果

流式消息中的图表块尚未完整时，不应向用户展示由中间残缺内容产生的图表解析或渲染错误。图表内容完整后应正常进入渲染并展示结果。

## 影响范围

- React Native 流式 Markdown 渲染
- Mermaid
- ECharts
- Vega / Vega-Lite
- PlantUML
- DOT / Graphviz
- 其他通过 fenced code block 接入的图表引擎

普通非流式、一次性传入完整 Markdown 的场景不受影响。

boomsi:
## Summary

- preserve explicit diagram fence closure in AST v2 across Web and native parsers
- defer Web and React Native diagram engine work while a streamed fence remains open
- commit parsed AST, source state, render output, and highlighting as version-consistent snapshots
- distinguish `Receiving diagram (...)…` from actual `Rendering diagram (...)…`
- keep CommonMark EOF auto-close behavior once the source is complete

## Root cause

The parser auto-closes an open fenced block at EOF and previously discarded whether a real closing fence was present. Renderers therefore invoked diagram engines with partial streamed code, producing transient errors before the diagram became valid.

An additional async race could pair the latest `complete` prop with the previous partial AST while the replacement Markdown was still parsing. Fast JSON-based engines could reject that stale code before the new AST arrived. Parsed content and its source state are now committed together, so renderers never observe a cross-version combination.

## Behavior

- open fence + streaming source: show `Receiving diagram (...)…` and do not invoke the engine
- closed fence + streaming source: render immediately
- completed source: render even when the fence was auto-closed at EOF
- source version change: keep the previous internally consistent snapshot until the replacement is ready
- stable engine work: show `Rendering diagram (...)…`; genuine failures still show `Diagram error`

This applies uniformly to every diagram engine routed through the shared diagram node.

## Validation

- [x] Web async snapshot regression tests (4 cases): cross-version promotion, EOF auto-close, stale failure isolation, and stable engine errors
- [x] Core diagram state matrix (4 cases)
- [x] strict TypeScript check for the Web regression test
- [x] `cargo test -p supramark-markdown` (54 unit tests + 26 public API tests)
- [x] native FFI fence-state test
- [x] `bun run test:core` (88 tests)
- [x] `bun run --filter @supramark/rn test` (41 tests)
- [x] `bun run --filter @supramark/web test` (4 tests)
- [x] core, React Native, and Web builds
- [x] targeted ESLint and formatting checks
- [x] iOS native bridge XCTest (4 tests)
- [x] Android JNI host bridge test
- [x] iOS and Android native artifact builds for all configured architectures
- [x] Loom TypeScript check and test suite (98 tests)

Closes #85


kookyleo (approved):
## Re-review: both blocking items fixed, most nits addressed too — approving

Against the previous round:

1. ✅ `DiagramBlock.tsx:23-26` restores the raw-code fallback under the "rendering" placeholder — if `collectRenderTasks` ever diverges from `renderNode`'s reachable set, the worst case is placeholder + full source, no more silent wedge. Streaming behavior is unaffected (the receiving state returns earlier in WebDiagramNode; verified the fallback cannot leak partial streamed code).
2. ✅ RN defer/snapshot logic now has real tests: `__tests__/DiagramNode.test.tsx` drives the actual `DiagramNode` with a controlled engine mock and asserts zero engine calls while `streaming && !fence_closed`, immediate render for closed fences, and the full receiving→rendering→svg transition with exactly one engine call. 44 pass.

Beyond the asks: legacy `syntax/fence.ts` deleted (the hardcoded `fence_closed: true` path is gone); `#[serde(default)]` added with a legacy-AST deserialization test and ast-spec.md updated; RN gains testIDs mirroring web's data attributes; the container-terminated-fence conservatism is now documented at the decision site.

On `f9b1cff3` (bob build skip) — verified it is a necessary consequence of this PR, not scope creep: adding `react-test-renderer` re-hoists `bob` to the workspace root, which flips the old `command -v bob` gate for four RN shell packages that have no bob config; counterfactually, `bunx bob build` in one of them fails with "No configuration found", i.e. the root build would break without this fix. The new gate reads the package's own `react-native-builder-bob` field and fails safe.

Local verification: core 88 pass (wasm rebuilt from PR code), web 4 pass, rn 44 pass, supramark-markdown `cargo test --lib` 55 pass including the new serde test; single-package builds clean. CI is fully green including the previously flaky type-check job.

### 🟡 Follow-ups (non-blocking)

- The restored code fallback itself has no direct assertion (web tests never reach the rendering state thanks to snapshot retention) — a unit test rendering `DiagramBlock` with `result={undefined}` would pin the exact branch that regressed last time.
- `react-test-renderer` is deprecated upstream; fine on React 19, worth revisiting on the next React bump.
- Placeholder strings went from hardcoded English to hardcoded Chinese — consistent with existing error text, but still no customization hook for hosts.
- CLAUDE.md still lists `src/syntax/fence.ts`, now deleted; `bun.lock` lost its `configVersion` line (bun version drift, cosmetic).

Streamed still-open `$$` math remains the known out-of-scope sibling issue — worth its own follow-up.

PR #88, opened 2026-07-21
refactor(rn): return canvas background ownership to host

boomsi:
## Background

Follow-up to #83. In #83 the gap-fix silently commented out `darkThemeStyles.root.backgroundColor` / `lightThemeStyles.root.backgroundColor`; review flagged it as an unrelated, undisclosed change. #83 reverted those two lines to stay scoped. This PR addresses the underlying question that comment raised: **should the renderer own its canvas background at all?**

## Root cause

The renderer's job is content — text, headings, code blocks, tables. The opaque canvas behind it belongs to the host (chat bubble, card, page). Painting `#0d1117` / `#ffffff` from inside `root`:

- Forces a hard rectangle that clashes with every host container (a chat bubble's rounded corners and brand color get covered by a black/white slab).
- Masks a design gap: the `dark` theme ships light foreground text (`#e0e0e0`) but left the matching dark canvas to guessing — `theme="dark"` without a dark host = unreadable.

## Approach: return the canvas to the host, keep content colors theme-driven

Split "colors" into two layers with different ownership:

| Layer | Owner | Examples |
|---|---|---|
| Canvas background (`root`) | **Host** (this PR removes it) | the opaque page / bubble / card behind content |
| Content colors (foreground + element decoration) | **Library, theme-driven** (unchanged) | text color, code-block bg, borders, map-card bg |

Only the root canvas is returned to the host. Foreground and per-element colors stay in `darkThemeStyles` — that is the library's job (making content readable on a given canvas).

## Changes

- **`styles.ts`**: drop `darkThemeStyles.root`; remove `lightThemeStyles` (it only ever added a white canvas, identical to the default — `theme="light"` now means "default foreground"). Export `themeBackground = { light: '#ffffff', dark: '#0d1117' }` — the recommended host canvas color paired with each built-in theme's foreground.
- **`Supramark.tsx`**: merge logic simplified (`theme === 'dark' ? darkThemeStyles : undefined`); the `theme` prop now documents the contract — the host must provide a canvas whose light/dark matches the theme, and `themeBackground` is the recommended value.
- **`App.tsx`**: the example host paints its own canvas via `themeBackground` — light mode gains an explicit canvas it previously borrowed from the component; dark mode switches from a hardcoded `#0d1117` to the export. This demonstrates the new contract.
- **`__tests__/styles.test.ts`** (new): locks the contract — `darkThemeStyles.root` undefined, `lightThemeStyles` removed, `themeBackground` values, `mergeStyles` behavior.

## Relationship to #83

#83 is a gap / trailing-margin fix and should stay scoped to that. This PR is the standalone, documented removal of canvas ownership. They are independent and land in either order.

## Test plan

- [x] `bun test` in `@supramark/rn` — 7 pass (new `styles.test.ts`)
- [x] `tsc --noEmit` in `@supramark/rn` — clean
- [ ] Simulator, light mode: detail screen renders on an explicit white canvas (previously transparent, borrowed from the component)
- [ ] Simulator, dark mode: visually unchanged (canvas color identical, now sourced from `themeBackground`)

### DIFF
--- a/examples/react-native/App.tsx
+++ b/examples/react-native/App.tsx
@@ -29,7 +29,7 @@ import {
   NativeModules,
 } from 'react-native';
 
-import { Supramark } from '@supramark/rn';
+import { Supramark, themeBackground } from '@supramark/rn';
 import type { SupramarkConfig } from '@supramark/core';
 
 // Feature metadata — id/version is read off these for the FeatureConfig list.
@@ -167,7 +167,7 @@ export default function App() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const containerStyle = [styles.container, isDark && { backgroundColor: '#0d1117' }];
+  const containerStyle = [styles.container, isDark && { backgroundColor: themeBackground.dark }];
   const headerStyle = [styles.header, isDark && { borderBottomColor: '#30363d' }];
   const titleStyle = [styles.title, isDark && { color: '#ffffff' }];
   const subtitleStyle = [styles.subtitle, isDark && { color: '#8b949e' }];
@@ -225,7 +225,10 @@ export default function App() {
     );
   }
 
-  const detailContentStyle = [styles.detailContent, isDark && { backgroundColor: '#0d1117' }];
+  const detailContentStyle = [
+    styles.detailContent,
+    { backgroundColor: themeBackground[isDark ? 'dark' : 'light'] },
+  ];
   const demoTitleStyle = [styles.demoTitle, isDark && { color: '#ffffff' }];
   const demoDescriptionStyle = [styles.demoDescription, isDark && { color: '#8b949e' }];
   const demoSectionTitleStyle = [styles.demoSectionTitle, isDark && { color: '#ffffff' }];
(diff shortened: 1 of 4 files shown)


kookyleo (approved):
## Review: SAFE TO MERGE — clean ownership split, contract locked by tests, zero-conflict with #83 either way

Verified against source: only the root canvas leaves the library — `darkThemeStyles` keeps all foreground/element colors (text, code-block bg, table header, map-card set), locked by the new `__tests__/styles.test.ts`; `lightThemeStyles` had zero consumers repo-wide outside this PR's own files; `themeBackground` exports through the existing `export *`; the example app now demonstrates the host-owned canvas on both pages. `theme='light'` ≡ no theme is documented in the prop's JSDoc. This also aligns RN with the web renderer's existing model, where the host has always owned the canvas.

Merge interaction with #83 tested locally in both orders from current main: zero conflicts, and the two resulting trees are identical (83's background restore is net-zero against base, so it cannot resurrect what this PR removes). Merged-tree rn suite: 54 pass / 0 fail, tsc clean. No ordering constraint.

🟡 Non-blocking: the canvas contract lives only in JSDoc + `themeBackground` today — `docs/guide/getting-started.md` shows `@supramark/rn` usage without mentioning `theme` or the host-canvas requirement; worth a short section there.

Not verified locally: real-device look of the dark example pages (transparent root over host canvas) — one simulator pass recommended whenever #83's spacing check happens.

PR #90, opened 2026-07-21
test(rn): register shared react-native mock to fix cross-file clobbering

kookyleo:
## Summary

bun's `mock.module` registry is process-wide. `spacing.test.ts` / `styles.test.ts` (from #83/#88) registered a StyleSheet-only `react-native` mock while `DiagramNode.test.tsx` (from #86) registered a wider surface. The later registration clobbers the earlier one, and already-loaded modules can then fail to resolve `ActivityIndicator`:

```
SyntaxError: Export named 'ActivityIndicator' not found in module '.../react-native/index.js'.
```

Each suite landed green from its own PR — the CI runs never had both files in one process. After the merges the failure is **environment-dependent**: it reproduces locally on bun 1.3.5 and the pinned 1.3.14 alike (both direct `bun test` and the CI-style `bun run --filter '@supramark/rn' test` invocation; the module error aborts the file and two tests never run), while the GitHub runner happens to pass — so green CI does not cover it.

## Fix

Register one full-surface react-native mock in `__tests__/support/mock-react-native.ts` and import it from all three test files instead of per-file `mock.module('react-native', ...)` calls. The suite no longer depends on mock registration order.

## Validation

- `bun test` in `packages/renderers/rn`: 57 pass / 0 fail (was 55-ran / 1 fail / 1 error), stable across 3 consecutive runs
- `bunx tsc --noEmit` and eslint clean on the touched files
- Verified pre-fix failure and post-fix pass under both bun 1.3.5 and 1.3.14

### DIFF
--- a/packages/renderers/rn/__tests__/DiagramNode.test.tsx
+++ b/packages/renderers/rn/__tests__/DiagramNode.test.tsx
@@ -4,6 +4,8 @@ import { create, act, type ReactTestRenderer } from 'react-test-renderer';
 import type { DiagramRenderResult } from '@supramark/engines';
 import type { SupramarkDiagramNode, SupramarkSourceState } from '@supramark/core';
 
+import './support/mock-react-native';
+
 // react-test-renderer 需要显式开启 act 环境，否则 effect 不会同步 flush。
 (globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT: boolean }).IS_REACT_ACT_ENVIRONMENT =
   true;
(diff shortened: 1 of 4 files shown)


boomsi (commented):
@kookyleo

Re-review: verified — looks good to merge

Verified `@supramark/rn` at HEAD `37dadbb3`:

✅ Reproduced on `main` (`bun test` in `packages/renderers/rn`): `55 ran / 1 fail / 1 error` — `SyntaxError: Export named 'ActivityIndicator' not found in '.../react-native/index.js'`; `DiagramNode.test.tsx` aborts on the module error and two cases never run. Matches the body.
✅ PR branch: `57 pass / 0 fail` across 4 files (+2 recovered), stable on repeat. Local bun is `1.2.21` (body cites 1.3.5/1.3.14) — still reproduces, which only reinforces the "env-dependent but version-broad" claim.
✅ eslint on the 4 touched files: exit 0, `0 problems`.

Note (non-blocking): `packages/renderers/rn/tsconfig.json:6` has `include: ["src", "../../core/src/**/*.d.ts"]` — `__tests__` is excluded, so the body's "tsc --noEmit clean on the touched files" is technically a no-op: tsc never compiles any touched file (all under `__tests__/`), so the exit 0 is an empty run. Not a blocker — the changes are minimal (drop inline `mock.module`, add one `import`, add one full-surface mock) and are covered by eslint + the running suite; the tsconfig gap is pre-existing, not introduced here. Suggest dropping the tsc line from the body or rewording to "test files aren't in tsc scope; covered by eslint + bun test".

Tests: rn suite — main `55 ran / 1 fail / 1 error` → branch `57 pass / 0 fail / 0 error`; eslint `0 problems` on touched files.

Ready to merge.

boomsi (approved):

Issue #95, opened 2026-07-24
RN renderer: list items wrap prematurely under streaming; ordered lists lose numbering


boomsi:
## Summary

In the React Native renderer, `list_item` nodes wrap their text prematurely (as if the available width were only ~4 CJK characters) when the markdown is fed incrementally (streaming/typewriter). Plain paragraphs in the same document wrap normally. Ordered lists additionally lose their numbering — every item shows a `•` regardless of `ordered`/`start`.

## Reproduce

Render this through `Supramark` (RN) with content growing character-by-character (typewriter re-parses on each tick) and `LayoutAnimation.configureNext` fired on line breaks:

```
使用说明
1. 输入 help → 输出本说明
2. 输入数字大于100时 → 生成 6~9 段文字，每段间隔2秒返回
3. 输入数字1~23时 → 即时返回不同类别的supermark格式
4. 输入其他字符时 → 即时原样返回
```

List items wrap every ~4 characters; paragraphs wrap to full width. Changing `1.` to `1,` (so it parses as a paragraph) makes it wrap correctly — confirming the issue is isolated to the `list`/`list_item` render branch.

## Root cause

`packages/renderers/rn/src/Supramark.tsx`, `case 'list_item'` renders:

```jsx
<View style={styles.listItem}>              // flexDirection: 'row'
  <Text style={styles.bullet}>•</Text>
  <Text style={styles.listItemText}>         // flex: 1
    {renderInlineNodes(item.children, ...)}
  </Text>
</View>
```

A `flex:1` `<Text>` inside a `row` `<View>` relies on yoga flex distribution + text measure to determine its wrapping width. Under streaming re-render (each tick re-parses and re-renders a growing markdown substring) combined with `LayoutAnimation.configureNext` on line breaks, this measure is unstable and the computed width collapses, causing premature wrapping.

A plain `paragraph` is a single `<Text>` as a direct child of a column container, which stretches to the parent width reliably — so paragraphs are unaffected.

## Secondary bug

The bullet is hardcoded to `'•'`:

```jsx
<Text style={styles.bullet}>{isTaskList ? checkSymbol : '•'}</Text>
```

`list.ordered` / `list.start` (see `SupramarkListNode`, `packages/core/src/ast.ts`) are ignored, so ordered lists render as unordered bullets.

## Suggested fix

[...]

PR #92, opened 2026-07-22
perf(rn): cache render results across virtual-list remounts

boomsi:
## Summary

- cache completed parsed documents and normalized diagram SVG in bounded renderer-level LRU caches
- restore cached content synchronously across virtual-list remounts and deduplicate equivalent in-flight work
- treat `options.cache` as the global diagram-cache fallback while preserving diagram and engine overrides
- share cache entries across equivalent inline config objects without relying on config reference identity
- export the core cache utilities from the React Native entry and add policy, remount, streaming, failure, and normalization coverage

## Why

Virtualized React Native lists unmount and remount message rows. Previously, remounting could repeat Markdown parsing, code highlighting, diagram rendering, and SVG normalization, exposing raw Markdown or a diagram placeholder before the final content appeared. Those intermediate height changes caused visible flashing and scroll-position jitter.

The first cache implementation also scoped entries by config object identity, so common inline config literals silently missed the cache. This change keys bounded renderer-level caches by the inputs that affect output and documents the current config-independent parse contract.

## Impact

Completed content can render from cache on the first remount frame. Streaming documents remain uncached, failed work is retried instead of retained, and explicit diagram or engine cache policies continue to override the global default.

Fixes #91.

## Test plan

- [x] `bun test packages/renderers/rn/__tests__` — 73 passed
- [x] `bun run --filter @supramark/rn build`
- [x] `bun run --filter @supramark/core build`

PR #101, opened 2026-07-24
fix(rn): list_item wraps prematurely under streaming; ordered lists lose numbering

boomsi:
## Summary
- `list_item` wraps its text prematurely (every few characters) when markdown is fed incrementally (streaming/typewriter) + `LayoutAnimation`. Paragraphs in the same document wrap fine.
- Ordered lists lose their numbering — every item shows `•` regardless of `ordered`/`start`.

Refs #95. (Reference rather than auto-close: ordered task lists still lose their number — see review.)

## Root cause
`list_item` rendered as `<View flexDirection:row><Text>•</Text><Text flex:1>…</Text></View>`. A `flex:1` `<Text>` inside a row `<View>` relies on yoga flex distribution + text measure for its wrapping width; under streaming re-render (each tick re-parses a growing substring) + `LayoutAnimation.configureNext` on line breaks, this measure is unstable and the width collapses → premature wrapping. A `paragraph` is a single `<Text>` directly under a column, which stretches to the parent width reliably, so it is unaffected.

Secondary: the bullet was hardcoded to `'•'`; `list.ordered` / `list.start` were ignored.

## Fix
- Render `list_item` as a plain `<Text style={paragraph}>` with the marker prefixed as a string — drops the row + `flex:1` measure path entirely (root-cause fix, not a `minWidth:0` patch).
- `case 'list'` computes the ordered marker (`${start + index}.`) or `•` and passes it to `list_item` via a new optional `listMarker` param on `renderNode`.

## Trade-off
Wrapped continuation lines no longer hanging-indent under the marker (they start at the left edge). Acceptable for most lists.

## Test plan
- [x] `tsc --noEmit` (packages/renderers/rn) — exit 0
- [x] `bun test` (packages/renderers/rn) — 61 pass, 0 fail (incl. new `list.test.tsx`)
- [x] `bun test` (packages/renderers/rn-selection) — 145 pass (selection traverses AST, unaffected)
- [x] Verified on device: streaming `1. … 2. … 3. … 4. …` now wraps to full width and shows numeric markers
- [x] Added `list.test.tsx` asserting marker output (ordered / unordered / task / `start`). `toContain("1. a")` only holds when marker + content share one `<Text>`, so it also pins the plain-Text shape — the premature-wrapping regression can't silently return (the pre-existing spacing test only asserted static styles and stayed green while rendering was actually broken)
- [x] Extracted shared `react-native-svg` + `@supramark/engines/rn` mocks into `support/mock-renderer.ts` (bun's `mock.module` registry is process-wide; a per-file engine mock clobbered `DiagramNode.test`'s controlled engine)


kookyleo (approved):
Approving per the review above.

@boomsi — heads up, I pushed one merge commit to this branch (`ff45f78f`). #92 landed first and conflicted with it in `__tests__/DiagramNode.test.tsx`, plus git silently auto-merged `support/mock-react-native.ts` into an object literal with duplicate `ScrollView` / `TouchableOpacity` / `Linking` keys. I resolved both files in favour of main, because #92 replaced the process-wide `@supramark/engines/rn` mock with `diagramEngine` prop injection, which supersedes the `support/mock-renderer` indirection for that one test. `list.test.tsx` still imports `support/mock-renderer` for its own engine + svg mocks, unchanged. Verified locally: `bun test` 86/0, `tsc --noEmit` exit 0, eslint clean.

On the `Fixes #95` in commit `4e21ccd8`: merging this will auto-close #95, but ordered **task** lists still drop their number (`1. [ ] todo` → `☐ todo`, `Supramark.tsx:337`). I will reopen #95 right after the merge, scoped to that remaining case.

Issue #104, opened 2026-07-24
RN renderer: loose / nested list item content is silently dropped


boomsi:
## Summary
In the RN renderer, list items whose `children` contain **block** nodes lose their content:
- **loose list** (blank line between items): each item's body disappears — only the marker (`•` / `1.`) remains.
- **nested list**: the parent item's own text renders, but the nested sub-list does not.

Tight lists (no blank lines, no nesting) are unaffected.

## Reproduce
Render through `Supramark` (RN):

loose:
```
1. 第一项

2. 第二项

3. 第三项
```
Expected: three numbered items with text. Actual: only `1.` `2.` `3.` show; the item bodies are blank.

nested:
```
- 外层
  - 内层A
  - 内层B
```
Actual: `外层` shows; the `内层A` / `内层B` sub-list is missing.

## Root cause
`list_item` renders its content via `renderInlineNodes(item.children)` (`packages/renderers/rn/src/Supramark.tsx`, `case 'list_item'`). `renderInlineNode`'s switch only handles inline node types and falls through to `default → null` for block nodes.

Parser output (verified by running `parse` directly):
- tight `- a\n- b` → `list_item.children = [text]` (inline) ✓
- loose `- a\n\n- b` → `list_item.children = [paragraph]` (block) ✗ → null
- nested `- outer\n  - inner` → `list_item.children = [text(outer), list]` ✗ → list dropped

## When introduced
`ea805a07` (2026-06-24, "适配 AST v2"). That commit switched `admonition` and `footnote` from `renderInlineNodes` to `renderNode` to handle block children, and fixed the **web** `list_item` inline case — but the **RN** `list_item` was missed and still calls `renderInlineNodes`, so block children (`paragraph`, nested `list`) fall through to `null`.

## Suggested fix
Render `list_item` block children with `renderNode` (as `footnote` / `admonition` already do — see `Supramark.tsx:558`). Caveat: must preserve the width fix from #101 — the item must not regress to the `row + flex:1` shape that caused premature wrapping. A column `<View>` (under which `<Text>` stretches to full width, unlike `flex:1` under a row) with the marker prefixed onto the first inline run handles tight / loose / nested uniformly; tight lists can keep the plain `<Text>` shape from #101.

boomsi:
[...]

PR #105, opened 2026-07-24
fix(rn): render loose/nested list item block children instead of dropping them

### LINKED ISSUE #103: RN renderer: loose list items render blank, nested sub-lists are dropped
## Summary
In the RN renderer, **loose lists (items separated by blank lines) render each item's body blank** — only the marker (`•` / `1.`) shows. **Nested sub-lists are dropped entirely** (the parent item's own text renders, but the indented child list disappears). Tight lists render correctly.

## Reproduce
Render through `Supramark` (RN):
```
1. 第一项

2. 第二项
```
→ only `1.` `2.` markers; body text「第一项…」missing.

```
- outer
  - innerA
  - innerB
```
→ `outer` renders; `innerA`/`innerB` disappear.

Removing the blank lines (tight) renders correctly — confirms it's structural, not content.

## Root cause
`packages/renderers/rn/src/Supramark.tsx`, `case 'list_item'` calls `renderInlineNodes(item.children)`. `renderInlineNode`'s switch has no `paragraph` / `list` case → `default → null`, so any **block** child of `list_item` is dropped.

Verified by parsing with the core parser (`parse`):

| input | `list_item.children` |
|---|---|
| tight `- a\n- b` | `[text]` ✅ |
| loose `- a\n\n- b` | `[paragraph]` ❌ dropped |
| loose ordered `1. a\n\n2. b` | `[paragraph]` ❌ dropped |
| nested `- outer\n  - inner` | `[text, list]` ❌ `list` dropped |

## When introduced
`ea805a07` (2026-06-24, "适配 AST v2 opaque container"). AST v2 / Rust parser makes loose `list_item.children = [paragraph]` and nests `[list]` (block). That same commit switched RN admonition and footnote from `renderInlineNodes` to `renderNode` to handle block children, **but the RN `list_item` was missed** (only the web `list_item` inline-node case was fixed). So the regression has been present since AST v2 adoption.

Note: the parser does not populate the `tight` field (outputs `undefined`), so the renderer can't branch on `item.tight` — it must handle block children unconditionally.

## Suggested fix
Render `list_item` as a column `<View>` and recurse via `renderNode` (like footnote at `Supramark.tsx:558`), with the marker prefixed on the leading inline run so the first line still wraps as a single `<Text>`. Must stay width-stable (no row + `flex:1`) — needs to reconcile with the plain-`<Text>` fix in #101.

## Related
- #95 / #101 — tight list premature wrapping (different bug, same `list_item` case).

boomsi:
Fixes #103.

> Note: #104 was a duplicate of #103 (same bug) and has been closed; this PR now targets #103.

## ⚠️ Stacked on #101
This branch is based on #101 (the list_item width fix), which is not merged yet. Until it lands, the diff below also includes #101's commits; GitHub hides them automatically once #101 merges, and I'll rebase this onto main at that point.

## Summary
- **loose list** (blank line between items): item bodies were silently dropped — only the marker (`•` / `1.`) survived.
- **nested list**: the sub-list was not rendered at all.

Tight lists are unaffected.

## Root cause
`list_item` rendered its content via `renderInlineNodes`, whose switch only handles inline node types and falls through to `default → null` for block nodes. Under AST v2 the parser produces `[paragraph]` for loose lists and `[text, list]` for nested lists, so the paragraph body and the nested sub-list were dropped.

Introduced in `ea805a07` (AST v2 adaptation), which switched `admonition` / `footnote` from `renderInlineNodes` to `renderNode` to handle block children — but missed the **RN** `list_item`.

## Fix
- Tight lists (inline-only children): unchanged — keep the plain `<Text>` shape from #101 (width-safe).
- When block children are present (loose / nested): render inside a column `<View>` (`listItemBlock`) via `renderNode`, so `paragraph` and nested `list` are no longer dropped. Under a column, `<Text>` stretches to full width, so this does **not** reintroduce the `row + flex:1` measure path that caused premature wrapping (#101).
- The marker is prefixed onto the first inline run, or onto the first `paragraph` when the item starts with a block.

## Test plan
- [x] `tsc --noEmit` (packages/renderers/rn) — exit 0
- [x] `bun test` (packages/renderers/rn) — 64 pass, 0 fail (incl. new loose/nested cases)
- [x] Added `list.test.tsx` cases: loose unordered (`• a`), loose ordered (`1. a`), nested (`• outer` + `• inner`)
- [x] Verified on device: loose and nested lists now render their content and sub-lists


kookyleo (dismissed):
Approving. Checked the two things most likely to be wrong in a refactor of this shape:

- **`INLINE_NODE_TYPES` is complete.** The set lists 10 types; `renderInlineNode`'s switch handles exactly those 10 (`text`, `strong`, `emphasis`, `inline_code`, `math_inline`, `link`, `image`, `break`, `delete`, `footnote_reference`). No inline type falls through to the block path. The comment asking future edits to keep the two in sync is the right guard for a hazard that has no compiler check.
- **The width fix is preserved.** Routing tight items through a plain `<Text>` and block items through `flexDirection: 'column'` both avoid the `row + flex:1` shape that caused the premature wrapping in #101, so this does not regress that fix while fixing #103.

One gap, which is pre-existing rather than introduced here:

**Ordered task lists still lose their number.** `packages/renderers/rn/src/Supramark.tsx`:

```ts
const marker = isTaskList ? `${checkSymbol} ` : `${listMarker ?? '•'} `;
```

`listMarker` is still discarded whenever `checked !== undefined`, so `1. [ ] todo` renders as `☐ todo`. That is exactly the item left open when #95 was reopened, and the new `list.test.tsx` does not cover it — its task-list case only uses `ordered: false`, so the gap is invisible to the suite. The shape suggested on #95 still applies:

```ts
const marker = `${listMarker ?? '•'} ${isTaskList ? checkSymbol + ' ' : ''}`;
```

Not blocking this PR, since #105's scope is loose/nested block children and it does that correctly. Leaving it on #95.

kookyleo (approved):
Re-approving after the branch was updated onto current main.

The update surfaced a real integration break that neither side showed alone: #92 narrowed `highlighted` to `ReadonlyMap` across the RN render chain after this branch was cut, and `renderListItemBody` was the only one of the six declarations still asking for a mutable `Map`, so `renderNode`'s call site stopped type-checking. Pushed `b4e41680` to fix it — a type-only widening; the helper only reads the map.

Verified on the merged tree before pushing: `tsc -p packages/renderers/rn/tsconfig.json --noEmit` clean, `bun test` 90 pass / 0 fail across 7 files. CI on the updated branch is 13/13 green.

The earlier review notes stand: `INLINE_NODE_TYPES` covers exactly the 10 types `renderInlineNode` handles, and the tight/block split preserves the #101 width fix. The ordered-task-list gap stays on #95.

Issue #106, opened 2026-07-24
Harden CommonMark conformance harness: regression-gated exits, PR trigger, comparison blind spots


kookyleo:
Follow-ups from the #102 review. All items are in the test harness; none affect production code. Baseline numbers were independently reproduced locally (semantic 181 / visual 124, exact case-ID match, zero drift).

### 1. Conformance run fails on any failure instead of gating on regressions
`tests/markdown-conformance/scripts/run-commonmark.mjs` drives its exit code from the absolute failure count, while `compareWithBaseline` already computes `added`/`resolved` but only feeds the report. With 181 + 124 recorded baseline failures, every push touching the workflow paths goes red on main and churns the tracking issue, yet a PR introducing new regressions is never blocked (the workflow does not run on pull requests).

Suggested: exit non-zero only when `baseline.overall.added > 0` (keep an env switch for absolute mode on workflow_dispatch), and run the semantic-only mode on pull requests so regressions are caught before merge.

### 2. Scale-only size-delta exemption can mask diagram collapse
`tests/official-diagram-visual/scripts/tools/official-diagram-visual-workflow.mjs` — `hasScaleOnlySizeDelta` keeps a case as `pass` even at -96% area when width/height shrink proportionally and the pixel band passes; the self-test was repurposed from "downgrades severe size collapse to review" to "keeps scale-only size collapse as pass". Combined with the new 540px default preview width, this can hide "diagram renders tiny" regressions.

Suggested: keep `review` for severe area collapse, or bound the exemption to moderate area deltas.

## Nits

- `lib/semantic/html-semantics.mjs:163` — operator precedence makes the guard `pre || (code && has-newline)`, so whitespace runs inside inline code spans are collapsed before comparison. Verified currently masks nothing (failure count unchanged with the fix applied), but it is a blind spot for exactly the regressions this suite should catch. Fix: `if (parentTag === 'pre' || parentTag === 'code') return normalized;`
- Metadata claims `profile: "commonmark"` but the runner uses the default feature set (GFM + linkify), so some recorded failures (notably Autolinks) are intentional superset behavior rather than bugs; label them or expose a pure-CommonMark CLI profile.
- `.prettierignore` does not exclude `tests/cases/_fixtures/`; running `format` rewrites `cases.json` and breaks the CI reproducibility check (`git diff --exit-code`).
[...]

PR #118, opened 2026-07-24
ci: fix CommonMark issue label handling

alpacassarg:
## 变更说明

修复 CommonMark conformance 工作流无法自动创建或更新失败 Issue 的问题。

## 根因

工作流使用了 GitHub CLI 不支持的命令：

```bash
gh label view bug

### DIFF
--- a/.github/workflows/commonmark-conformance.yml
+++ b/.github/workflows/commonmark-conformance.yml
@@ -151,12 +151,26 @@ jobs:
             exit 1
           fi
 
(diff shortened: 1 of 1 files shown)


kookyleo (approved):
Approving per the review above: the split-and-create loop is correct. The two follow-ups (`xargs` trimming unsafe for labels containing quotes/backslashes, and `always()` → `!cancelled()`) are non-blocking and tracked separately. #106 stays open — the `pull_request` branch at :122 is still dead until that trigger is added.

PR #121, opened 2026-07-24
fix: reach 652/652 CommonMark 0.31.2 conformance (parser + web renderer)

mybee:
Closes the remaining CommonMark 0.31.2 conformance gaps for the `production-web-renderer-dom` target. The suite now passes **652/652** cases (semantic + visual comparison against the cmark reference HTML, with raw HTML enabled), up from 181 failures on `main`.

## Parser (Rust) — `crates/supramark-markdown`
- **autolink**: add `to_ast_v2` so `<https://…>` maps to a `Link` node instead of falling through to the legacy text/html path.
- **html_block**: keep the trailing newline of an HTML block in the raw value. (Paragraph interruption by closing tags already worked on main; the `can_terminate_paragraph` change here is a comment only. The functional change in this file is `value: self.content.clone()` retaining the trailing newline.)
- **html_block regexps**: update the HTML comment pattern to the CommonMark 0.30 → 0.31.2 definition, which makes `<!-->` and `<!--->` valid comments. (The previous pattern already compiled fine under the `regex` crate — no lookahead was involved; the change is the semantic upgrade, not a compilation fix.)
- **supramark**: unescape backslash/entity escapes in fenced code info strings.
- **utils**: treat all Unicode Symbol categories (Sm/Sc/Sk/So) as punctuation for emphasis flanking, matching CommonMark 0.31.2 (so `*£*`, `*€*`, etc. behave correctly).

## Web renderer — `packages/renderers/web`
React's component model maps one node to one closed DOM element, so it cannot express the raw HTML fragments whose final tree the browser's tree-construction stage owns. The new paths route those fragments through the browser instead of dropping them:
- **`RawHtml`**: injects a value through a `<template>` and splices the resulting nodes in place of a hidden placeholder (layout effect, before the host reads `innerHTML`). Tracks the inserted node range and restores the placeholder's slot on cleanup, so a re-render removes exactly what the previous run added instead of duplicating it (otherwise streaming markdown grows without bound). Real-browser only; SSR/non-DOM drops the value.
- **`renderRawNode`**: falls back to verbatim `RawHtml` when a raw value carries trailing markup (stray close tag, content after a balanced pair) that a host element's `innerHTML` context would mis-parse.
- **`mergeRawNodes`**: emits a balanced raw wrapping block-level children verbatim (cmark's block-boundary newlines survive), and absorbs an unclosed block-container open tag's following siblings into one `RawHtml` so the open container folds them as the reference HTML expects.
- **Paragraphs with inline raw HTML** are emitted verbatim as `<p>…</p>\n` via `RawHtml`, so an unclosed inline tag like `<a href="bar">` triggers active-formatting-element reconstruction at `</p>` and leaks into trailing whitespace exactly as cmark's output does.
- Inline/block serializers back the verbatim paths and carry the configured `classNames` (strong/em/code/a/img/del, with `classNames.code` merged into the `language-xxx` class); the renderer falls back to component rendering whenever a child cannot be statically serialized (math, tables, task-list items), so interactive content is unaffected.

### Raw HTML is opt-in, default off
Raw HTML passthrough is gated behind `options.allowDangerousHtml` on `SupramarkConfig` (alongside `cache`/`strict`), defaulting to `false`. When off, raw nodes are dropped — no `innerHTML`/`dangerouslySetInnerHTML` surface from untrusted markdown, matching the pre-raw-HTML behaviour so an upgrade never silently enables script execution. The conformance harness sets the flag `true` because CommonMark's expected HTML is defined with raw HTML passed through; the 652/652 figure is "652/652 with raw HTML enabled". `createConfigFromRegistry()` does not set the flag. Defaulting off (rather than sanitising by default) because the parser already rejects `javascript:`/`vbscript:`/`data:` link URLs at parse time, and a sanitiser would not match the spec's expected HTML anyway.

## Tests
- `packages/renderers/web/__tests__/commonmark-block-render.test.tsx`: raw-HTML tests opt into the flag; added regressions for raw dropped when the flag is off (block + inline) and `RawHtml` not duplicating output across re-renders. 25/25 pass.
- `cargo test -p supramark-markdown --lib` gates the parser AST v2 mapping fixes (autolink→Link, HTML block type-6 interruption, fence info unescaping, is_punct_char symbol categories). 59/59 pass.
- Full 652-case CommonMark 0.31.2 suite (semantic + visual, production-web-renderer-dom) — 652/652.
- `tests/markdown-conformance/baselines/commonmark.json` updated to 0 semantic / 0 visual failures so the ratchet is taut (a regression back toward 181 surfaces as new failures).

## Test plan
- [x] `cargo test -p supramark-markdown --lib` — 59/59 pass
- [x] `bun run test:core` — 88/88 pass
- [x] `bun --filter @supramark/web test` — 25/25 pass
- [x] `bun run lint` + `bun run lint:types` — clean
- [x] Full 652-case CommonMark conformance suite (semantic + visual, `VISUAL_COMPARE=1`) — 652/652

Refs #107.


kookyleo (approved):
Approving per the round-2 review: both blockers closed with measured evidence, and 652/652 semantic + 652/652 visual independently reproduced against real Chromium with the gate in place.

Merging now. The one remaining item — `generate-fix-report.mjs` still hardcoding an ID subset and the obsolete `181 → 87` header, referenced by no CI job or package script — is a cleanup, not a blocker; tracking it separately rather than holding conformance.

PR #122, opened 2026-07-24
fix(mermaid): reserve edge-label space at layout time for CJK (#93)

mybee:
## Summary

Refs #93 (supersedes the closed #94).

> **Note:** the `edge_label_decluster` flag is wired through the Rust crate and CLI only — it is not yet exposed to Supramark hosts via the `packages/web` / RN bindings (see review: wiring stops at the Rust crate boundary). The issue should stay open until that binding lands; this PR is `Refs #93`, not `Closes #93`.

Flowchart edge labels overlapped endpoint nodes and each other in dense `LR` layouts. **Root cause**: dagre reserves rank space for edge labels using the width the renderer measures, but the byte-exact jsdom shim under-measures CJK glyphs (~40% of real width) — e.g. "账户同步" measured ~28px vs ~64px real. So dagre under-reserved and the label box overflowed its endpoint nodes; the emitted `<foreignObject>` was also too narrow, so the visible text overflowed to one side and looked off-centre.

Adds an **opt-in** `edge_label_decluster` render option (off by default → default path stays byte-exact with `mermaid@11.14.0`). When enabled:
- **Layout**: the edge-label width is floored to a CJK-aware estimate (one EM per wide glyph via the Unicode East Asian Width property), so dagre reserves enough inter-node space for the label to fit between its endpoints. The floor runs over the **marker-free plain text** (tags/entities stripped), so `<br/>` / `**` / `&amp;` are not billed.
- **Render**: the emitted `<foreignObject>` uses that same width so the visible text is centred between its endpoints instead of overflowing a too-narrow box.
- **Post-process**: a render-time pass nudges apart any labels that still collide, resolving collisions in screen space (per-frame, so isolated-cluster labels with cluster-local coordinates are not mistakenly moved).

Supersedes #94 (pure post-process), which could not fix label/node overlaps where the label is wider than the gap.

## Verification

Chrome headless 1:1 measurement on the issue #93 repro:
- Default (flag off): `LvL=1 LvN=4` — unchanged, matches upstream `mermaid@11.14.0`.
- Decluster (flag on): `LvL=0 LvN=0`; "账户同步" glyph-ink centre offset from the Phone↔Laptop midpoint went **+20.6px (right-shifted) → -1.2px (centred)**.
- CJK label with a `<br/>` tag floors to ~4×EM (64), not the raw-markup ~108 — the over-wide box no longer re-creates the off-centre overlap.

## Test plan

- [x] `cargo test -p mermaid-little` — 1030+ pass / 0 fail (lib + examples + integration)
- [x] `cargo test --test edge_label_decluster` — opt-in AABB non-overlap, byte-exact default, CJK width floor, **markup-floor regression** (`<br/>` CJK label does not over-reserve)
- [x] `cargo test --test edge_label_frames` — isolated-cluster coordinate-frame correctness
- [x] `cargo test --test flowchart_inline` — default byte-exact snapshots unchanged
- [x] `cargo clippy --all-targets` — zero warnings on changed files
- [x] `cargo fmt --check` — clean
- [ ] CI green

## Known follow-ups (not blocking, per review)

- `is_cjk` → `is_wide_glyph` now uses `unicode_width` (covers Hangul, CJK Ext B+, Bopomofo; excludes halfwidth forms).
- `decluster_with_obstacles` is best-effort (does not guarantee zero overlap when `max_displacement` binds); the integration assertion holds for this fixture but the pass is documented as best-effort.
- The flag is not yet exposed through `packages/web` / RN — tracked as the reason this is `Refs #93` rather than `Closes #93`.
- CI has no `cargo test -p mermaid-little` job (d2-little / plantuml-little do) — worth a separate PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -875,6 +875,7 @@ dependencies = [
  "supramark-diagram-core",
  "supramark-font-metrics",
  "thiserror 1.0.69",
+ "unicode-width",
 ]
 
 [[package]]
--- a/crates/mermaid-little/Cargo.toml
+++ b/crates/mermaid-little/Cargo.toml
@@ -113,6 +113,10 @@ supramark-diagram-core = { path = "../supramark-diagram-core", version = "0.1" }
 font_metrics_core = { package = "supramark-font-metrics", path = "../font-metrics", version = "0.1" }
 thiserror = "1"
 log = "0.4"
+# East Asian Width lookup, used by `layout::label_metrics` to decide which
+# glyphs render ~1em wide. Same version d2-little already pins, so the
+# workspace resolves to a single copy.
+unicode-width = "0.1"
 env_logger = { version = "0.11", optional = true }
 # QuickJS bindings for the optional `katex` feature. Pinned to 0.10 (last
 # release compatible with rust-version = "1.82"; 0.11+ requires 1.87).
--- a/crates/mermaid-little/examples/fc_diff_probe.rs
+++ b/crates/mermaid-little/examples/fc_diff_probe.rs
@@ -92,7 +92,7 @@ fn main() {
             theme::apply_theme_variables(&mut theme, tv);
         }
         let l = match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
-            layout_fc::layout(&d, &theme)
+            layout_fc::layout(&d, &theme, false)
         })) {
             Ok(Ok(l)) => l,
             _ => {
--- a/crates/mermaid-little/src/engine.rs
+++ b/crates/mermaid-little/src/engine.rs
@@ -10,7 +10,9 @@
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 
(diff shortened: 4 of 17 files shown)


kookyleo (approved):
Approving per the round-2 review: the compile blocker is gone (`cargo test -p mermaid-little` 1030/0/9 across 28 targets, fmt clean) and the coordinate-frame blocker is fixed properly, with a regression suite that includes a vacuity guard and a full-document equality assertion.

Merging with three round-1 `should fix` items still open, each now tracked as a follow-up issue rather than held against this PR:

- the per-glyph width estimate still runs over the raw markup string, so a label with `<br/>` or `**` reserves ~70% too much width;
- `decluster_with_obstacles` does not guarantee separation (9 of 15 pairs still overlap under the render config with 6 co-located rects), while `tests/edge_label_decluster.rs:180` asserts zero — correct for this fixture, latent flake if the layout shifts;
- the wiring stops at the Rust crate boundary, so no Supramark host can enable this yet.

Thanks for already changing the body to `Refs #93` — #93 stays open until the binding lands, which is right.

Issue #126, opened 2026-07-27
[mermaid-little] Edge-label decluster follow-ups from the #122 review: overlap assertion, host wiring


kookyleo:
Carried over from the round-2 review on #122, which was merged with these open. All three were re-measured on the merged revision `1bf9d5a7`; the numbers below are from that run, not from round 1.

### 1. The width estimate is billed against the raw markup string

`crates/mermaid-little/src/render/svg_flowchart.rs:4896` and `crates/mermaid-little/src/layout/flowchart.rs:1787` feed the raw label into `cjk_aware_label_width`, so every markup character is billed 0.56 em:

| label | reserved width |
|---|---|
| 4 ideographs, plain | 64.0 (correct) |
| same + `<br/>` | 108.8 |
| same wrapped in `**` | 99.84 |
| same + `&amp;` | 108.8 |

`measure_html_markup_label` (via `strip_html_for_measurement`) and the browser both measure the *stripped* text. So for a CJK label containing a line break the reserved box is ~70% too wide — which re-creates the off-centre appearance #122 exists to fix, for exactly the labels it targets.

Fix is one call: feed `strip_html_for_measurement(&markdown_label_to_html(text))` into `cjk_aware_label_width`.

### 2. The decluster pass does not guarantee separation, but the test asserts it does

`crates/mermaid-little/src/layout/edge_label_decluster.rs` — `decluster_with_obstacles` terminates with labels still stacked when `max_displacement` binds. Measured with the exact config used at `svg_flowchart.rs:4981` (`gap = 4`, `max_displacement = 30`, `max_iters = 24`) and 6 co-located 100×40 rects: **9 of 15 pairs still overlap** after the pass.

(One thing did improve over round 1: no two labels land on identical centres any more.)

`tests/edge_label_decluster.rs:180` asserts `decl_overlaps.is_empty()`. That holds for the #93 fixture but locks in a property the algorithm does not provide, so it is a latent flake the moment that layout shifts.

Suggest asserting "overlap count strictly decreases, and is 0 for this fixture", and documenting the pass as best-effort in its module header (the header already hints at this; the test does not).

### 3. The option is unreachable from any Supramark host

`crates/mermaid-little/packages/web/src/lib.rs` exports only `convert` / `convert_with_id` / `version`. Nothing under `packages/` references `EngineRenderOptions`, `render_with_options` or `edge_label_decluster`. The RN binding is unchanged.

[...]

PR #128, opened 2026-07-27
fix(mermaid): wire edge_label_decluster to web; relax decluster test

mybee:
## Summary
- **Item 2 (#126):** The decluster integration test asserted zero overlaps — a property the pass does not guarantee when `max_displacement` binds (with the #122 config, 9 of 15 co-located pairs still overlap). Replaced with two assertions: "overlap count strictly decreases" (the real, drift-robust guarantee) plus a fixture-specific zero check for the #93 repro. Documented the pass as best-effort in the `edge_label_decluster` module header.
- **Item 3 (#126):** `packages/web/src/lib.rs` only exported `convert` / `convert_with_id` / `version`, so no Supramark host could enable the option — only the CLI. Added a `convert_with_options` wasm export and threaded `edgeLabelDecluster` through `@supramark/engines` and the diagram engine config. A web host now opts in via `config.diagram.engines.mermaid.edgeLabelDecluster`. Default path stays byte-exact; an older wasm bundle without the new export falls back to `convert(code)` with a warning.

Item 1 from #126 was already fixed in `112abd88` (the issue's correction comment confirms). RN wiring is intentionally out of scope here — it is a larger native surface (FFI + iOS/Android + TurboModule spec); #93 stays open until that binding lands.

## Test plan
- [x] `cargo test -p mermaid-little --test edge_label_decluster` — 4 pass
- [x] `cargo test -p mermaid-little --lib edge_label_decluster` — 6 pass
- [x] `cargo check -p mermaid-little-web --target wasm32-unknown-unknown` — new export compiles
- [x] `bun test __tests__` (engines, full suite) — 20 pass / 6 files (wiring assertions merged into `mermaid-label-centering.test.ts` to avoid bun:test cross-file `mock.module` contamination)
- [x] `bun scripts/lint-types.ts packages/engines packages/core crates/mermaid-little/packages/web` — 0 errors
- [x] `cargo fmt` / `prettier` / `eslint` — clean
- [ ] `bun run build:wasm` — produces the new `convert_with_options` entry (run before deploy; not exercised in this worktree)

### DIFF
--- a/crates/mermaid-little/packages/web/src/index.ts
+++ b/crates/mermaid-little/packages/web/src/index.ts
@@ -14,7 +14,12 @@
  * ```
  */
 
-// Re-export the raw wasm-bindgen API. `convert`, `convert_with_id`, and
-// `version` are the public functions; everything else (`__wbg_set_wasm`
-// etc.) stays internal to the generated JS.
-export { convert, convert_with_id as convertWithId, version } from './wasm/mermaid_little_web.js';
+// Re-export the raw wasm-bindgen API. `convert`, `convert_with_id`,
+// `convert_with_options`, and `version` are the public functions; everything
+// else (`__wbg_set_wasm` etc.) stays internal to the generated JS.
+export {
+  convert,
+  convert_with_id as convertWithId,
+  convert_with_options as convertWithOptions,
+  version,
+} from './wasm/mermaid_little_web.js';
--- a/crates/mermaid-little/packages/web/src/lib.rs
+++ b/crates/mermaid-little/packages/web/src/lib.rs
@@ -31,6 +31,24 @@ pub fn convert_with_id(mmd: &str, id: &str) -> Result<String, JsValue> {
     mermaid_little::convert_with_id(mmd, id).map_err(|e| JsValue::from_str(&e.to_string()))
 }
 
+/// Convert with an explicit diagram id and non-byte-exact readability
+/// tweaks, mirroring [`mermaid_little::convert_with_options`].
+///
+/// `edge_label_decluster` toggles the opt-in edge-label collision-avoidance
+/// pass (#93). With it off, the output matches [`convert_with_id`] for the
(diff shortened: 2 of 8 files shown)


kookyleo (approved):
Approving. Verified the wiring end-to-end rather than taking the description at its word:

- **The wasm export is real, not inert.** `crates/mermaid-little/packages/web/src/wasm/` tracks only the `.d.ts` placeholder — the glue `.js` is generated by `bun run build:wasm`, so `convert_with_options` will exist once the bundle is rebuilt. The `console.warn` fallback in `packages/engines/src/mermaid/index.ts` is defensive, not the live path.
- **The host opt-in claim holds.** I expected this PR to need a change in the renderers to plumb `config.diagram.engines.mermaid.edgeLabelDecluster` through, and it does not: both `buildDiagramRenderOptions` (`packages/renderers/web/src/Supramark.tsx:1833`) and `buildRenderOptions` (`packages/renderers/rn/src/DiagramNode.tsx:300`) end with a generic passthrough loop that copies every engine-config key except `enabled`/`timeoutMs`/`server`/`cache`. The new field reaches `renderMermaidSvg` automatically.
- **RN scoping is correct.** RN routes to the native binding, so the option is a no-op there; leaving #93 open until that lands is the right call.

One thing the title oversells, which does not block the merge:

**#126 item 2 is not actually closed by this PR.** The title says "relax decluster test", but `assert!(decl_overlaps.is_empty())` is still there at `tests/edge_label_decluster.rs` — the change *adds* a weaker "overlap count strictly decreases" assertion in front of it rather than replacing it. The module header and comments now frame zero-overlap as a fixture-specific fact rather than an algorithm contract, which is a defensible way to keep the #93 repro pinned. But the brittleness #126 item 2 was filed about survives verbatim: any layout drift that leaves one pair overlapping still turns this test red, and the reader now has an explicit comment telling them to edit the assertion when it happens. Tracking that on #126 rather than holding this PR.

PR #84, opened 2026-07-17
feat(rn-selection): document-level selection system for React Native

kookyleo:
## Summary

Adds `@supramark/rn-selection` — a Supramark-owned, AST-as-source-of-truth document selection system for React Native, replacing what native `<Text selectable>` cannot do: cross-block selection and feature-aware copy payloads.

Layered bottom-up:

- **Core model** — `linearizeForSelection` (AST v2 → flat unit stream; `unit.text` holds plain text only, markdown affixes live in payloads/syntax units), `resolveSelectionRange` (range → covered units, grapheme-safe splitting via `Intl.Segmenter` with surrogate-pair fallback), `serializeSelectionUnits` (plainText / markdown / source / html — both plain-text and markdown copies are lossless).
- **Tables** — compositional structural units: full-table selection copies as GFM table / TSV / HTML; partial selection strips scaffolding via shared `structuralGroup` ids and degrades to clean tab-separated text.
- **Native contract (milestone 2, TS side)** — `TextSegmentHandle` matching the vendored `selectable-rich-text` command+event API; `segmentAdapter` with pure segment-local ⇄ document offset mapping.
- **Coordinator (milestone 3 core)** — document-ordered block registry, hit-test geometry, external-store selection state; thin React wiring (`SelectionRoot`, `SelectableBlock`, `SelectionOverlay`, `useDocumentSelection`) with all logic in pure tested modules.
- **Event wiring + overlay** — per-block sinks close the "native events carry no nodeId" gap; menu actions serialize the selection and deliver payloads through a host `onCopy` callback (package stays clipboard-free); block-level overlay with vertical rect merging.
- **Vendored native primitive** — `boomsi/selectable-library` under `native/selectable-rich-text` with upstream credit preserved; upstream devDependencies trimmed (they pulled a second react-native + ~900 packages into the workspace).

The default `@supramark/rn` renderer is unchanged; nothing imports this package unless the host opts in.

## Verification

- 126 bun tests, `tsc --noEmit` clean (strict).
- **iOS simulator (iPhone 17 Pro, RN 0.81 New Arch)**: vendored pod autolinks and builds; programmatic cross-block selection paints a merged block-level highlight (uncovered blocks excluded); markdown copy reconstructs heading prefix / bold / emoji losslessly; clear removes the overlay. Exercised via the example app's self-driving `SelectionDemo` screen.
[...]

PR #129, opened 2026-07-28
ci: put integration tests on the PR-gating path

### LINKED ISSUE #123: CI: no job runs `cargo test -p mermaid-little`, so the crate can fail to compile with all checks green
## What is missing

There is no CI job that runs `cargo test -p mermaid-little`.

- `ci.yml:130` runs `bun test` inside `crates/mermaid-little/packages/react-native`;
- `ci.yml:187` runs the JNI host script;
- `d2-little` and `plantuml-little` each have a dedicated Rust job (`d2-little Rust tests`, `d2-little Rust (integration + fmt + clippy)`, `plantuml-little Rust tests`, `plantuml-little reference-SVG suite`).

`mermaid-little` — the largest of the three crates — has none.

## Why it matters, concretely

PR #122 shipped a revision where `crates/mermaid-little/examples/fc_diff_probe.rs:95` still called `layout_fc::layout` with 2 arguments after the signature had grown to 3:

```
error[E0061]: this function takes 3 arguments but 2 arguments were supplied
  --> crates/mermaid-little/examples/fc_diff_probe.rs:95:13
```

`cargo test -p mermaid-little` and `cargo clippy -p mermaid-little --all-targets` both failed to compile. **All 13 CI checks were green.** The error was only found by a reviewer building the crate by hand.

The PR's own test plan ran `--lib` plus two named `--test` targets, which skips `examples/` — an easy and understandable gap, and exactly the kind of thing CI exists to catch.

## Suggested shape

Mirror the `d2-little` jobs:

- `mermaid-little Rust tests` — `cargo test -p mermaid-little` (all targets, so `examples/` is compiled);
- fold `cargo fmt -p mermaid-little --check` and `cargo clippy -p mermaid-little --all-targets` into the same job or a sibling, matching `d2-little Rust (integration + fmt + clippy)`.

Note the suite currently takes ~3 minutes locally (1030 tests, two targets dominating at ~100 s and ~85 s), so it is a real but acceptable CI cost. Clippy currently reports 19 pre-existing warnings in this crate (12 × redundant reference in `format!`, 7 × `?` operator), so `-D warnings` would need those cleaned first, or the job starts non-blocking.

kookyleo:
Three crates or suites could break with every check green. #121 and #105 are the two worked examples, both from the last four days.

## What was missing

**`tests/public_api.rs` never ran on a PR.** `ci.yml` ran `cargo test -p supramark-markdown --lib`, so the contract suite for what `parse()` hands the renderers only ran inside `commonmark-conformance.yml` — a workflow with no `pull_request` trigger. #121 changed HTML-block line-terminator semantics without updating two assertions there. The PR was green, main went red, and because that step runs before the CLI build and the conformance comparison, **the CommonMark comparison was skipped on every run from #121 until #127 landed**. Fixed by dropping `--lib`.

**`mermaid-little` had no CI job at all (#123).** The Test & Build job compiles `mermaid-little-web` for wasm32 via `bun run build:wasm`, but never builds or runs the crate's own tests — a native build break or a layout regression lands unnoticed. The decluster work in #122 / #128 touches layout code whose only regression coverage lives in these suites. Added a dedicated job: no native deps, no submodules, full `cargo test -p mermaid-little` (703 lib + integration tests, ~3.5 min of execution, cached). `katex` / `cose_bilkent` stay off — they vendor QuickJS and belong in a separate tier.

**Conformance only ran post-merge.** Added a `pull_request` trigger with the same path filter as the push trigger. The issue-filing step already carries a `github.event_name != 'pull_request'` guard that was dead code until now, so PR runs report through the check and the uploaded artifacts and never open or update an issue. This is one of #106's asks.

## Verified locally before enabling the PR trigger

The concern with adding the trigger is turning every touching PR red. It does not:

```
CommonMark 语义对照：通过 652/652，未通过 0
CommonMark 视觉对照：通过 652/652，未通过 0
```

That is `run-commonmark-visual.mjs`, the path the workflow runs by default. The non-visual path is 640/652 — tracked on #106, along with that path's baseline ratchet being inert due to `comparisonTarget` mismatch. Neither is touched here.

Also verified: `cargo test -p mermaid-little` 703 pass / 0 fail; `cargo test -p supramark-markdown` clean on this branch; both workflow files parse.

## Known gap this does NOT close — needs an admin

Documented in a comment under `on:` in `ci.yml` rather than silently left out.

Jobs already run against `refs/pull/N/merge`, so the merge result is what gets tested. What is missing is a **re-run when the base moves**: a green check stays green after main advances underneath it.

#105 is the worked example. Cut 2026-07-24 and green. #92 then landed on main and narrowed `highlighted` to `ReadonlyMap` across the RN render chain; #105's own helper still declared a mutable `Map`, so the two together did not type-check — while #105 sat green, because its checks were computed against the pre-#92 main and never recomputed. It only surfaced when the branch was explicitly updated, and needed `b4e41680` to fix.

Closing that needs **"Require branches to be up to date before merging"** (`strict_required_status_checks_policy: true`) on the `main` ruleset. That is admin-only — write access cannot set it. @g65537 @Camel9527, this is the ask.

Note it will immediately require #84 (37 commits behind main) to update its branch before merging.

Closes #123
Refs #106

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,28 @@ on:
   pull_request:
     branches: [main, develop]
 
+# KNOWN GAP -- needs a repository setting, not a change in this file.
+#
+# On `pull_request`, actions/checkout already builds the PR merged into its
+# base (refs/pull/N/merge), so every job below runs against the merge result
+# rather than the PR head. What is missing is a re-run when the base moves:
+# a green check stays green after main advances underneath it, and nothing
+# re-evaluates it before merge.
+#
+# #105 is the worked example. It was cut on 2026-07-24 and went green. #92
+# then landed on main and narrowed `highlighted` to ReadonlyMap across the
+# RN render chain. #105's own helper still declared a mutable Map, so the
+# two together did not type-check -- while #105 sat green, because its
+# checks were computed against the pre-#92 main and never recomputed. The
+# break only surfaced when the branch was explicitly updated.
+#
+# The fix is the "Require branches to be up to date before merging" toggle
+# on the `main` ruleset (`strict_required_status_checks_policy: true`),
+# which forces a re-run against current main before the merge button
+# unlocks. That is admin-only -- write access cannot set it. Until an admin
+# turns it on, treat a long-lived green PR as unverified and update its
+# branch before merging.
+
 jobs:
   # License compliance — REUSE spec + cargo-deny over the Rust workspace.
   # The workspace-internal SPDX field check runs in the test job
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,12 +102,24 @@ jobs:
         with:
           targets: wasm32-unknown-unknown
 
-      - name: Run supramark-markdown lib tests
+      - name: Run supramark-markdown tests (lib + integration)
         # Unit tests for the canonical parser's AST v2 mapping (autolink → Link,
         # HTML block type-6 paragraph interruption, fence info unescaping,
-        # punctuation classification). Single matrix entry is enough.
+        # punctuation classification), plus tests/public_api.rs — the contract
+        # suite for what `parse()` hands the renderers.
+        #
+        # This used to be `--lib`, which left tests/public_api.rs running in
+        # exactly one place: the `cargo test -p supramark-markdown --test
+        # public_api` step of commonmark-conformance.yml, a workflow with no
+        # `pull_request` trigger. #121 changed HTML-block line-terminator
+        # semantics without updating two assertions there, so main merged red
+        # and stayed red for three pushes — and because that step runs before
+        # the CLI build and the conformance comparison, those were skipped on
+        # every run, so the CommonMark comparison never actually executed.
+        # Dropping `--lib` puts the contract suite on the PR-gating path where
+        # a break is caught before merge.
         if: matrix.node-version == '20.x'
-        run: cargo test -p supramark-markdown --lib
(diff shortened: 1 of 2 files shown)


mybee (dismissed):
Verified the three claims against `main`:

- `crates/supramark-markdown/tests/public_api.rs` exists, and `ci.yml` on main runs `cargo test -p supramark-markdown --lib` (line 110) — so the contract suite only ran inside `commonmark-conformance.yml`, which has no `pull_request` trigger. Dropping `--lib` puts it on the PR-gating path where #121-class breaks get caught before merge. ✓
- `mermaid-little` has no dedicated test job on main (only d2-little and plantuml-little do); the new `mermaid-little Rust tests` job closes that gap, full `cargo test -p mermaid-little` with `katex`/`cose_bilkent` correctly left out as a separate tier. ✓
- `commonmark-conformance.yml` on main triggers only on `push` + `workflow_dispatch`; the added `pull_request` trigger mirrors the push path filter, and the existing `github.event_name != 'pull_request'` guard on the issue-filing step makes the split clean. ✓

Diff is CI-only (94/-3, two workflow files), no source touched. The known gap — green checks going stale when the base advances underneath — is correctly scoped out as an admin-side ruleset toggle (`strict_required_status_checks_policy: true`), not something this PR can land. Worth tracking that separately so #84 (37 commits behind) gets updated before merge once that toggle is on.

All 15 checks green. Approving to unblock.

mybee (approved):

PR #131, opened 2026-07-28
fix(mermaid-little): clear 19 clippy warnings to unblock gating

mybee:
## Summary

Clears the 19 pre-existing clippy warnings in `mermaid-little` that **#123** flagged as the blocker for making the clippy step gating. Unblocks the `continue-on-error` in PR **#129**.

- 12 × `redundant reference in format!` — `svg_journey` / `svg_sankey` / `svg_requirement` / `svg_xychart`
- 7 × `?` operator rewrite — `gantt` / `requirement` / `sequence` / `state` / `timeline`

All applied via `cargo clippy --fix --lib -p mermaid-little`. The `?` rewrites are semantically equivalent: the trailing `else { return None; }` branch moves onto the final `else if`'s pattern as a `?`, so the no-match path still returns `None` exactly as before.

## Verification

- `cargo clippy -p mermaid-little --all-targets -- -D warnings` — clean
- `cargo test -p mermaid-little` — **1031 passed / 0 failed** (9 ignored)
- `cargo fmt -p mermaid-little --check` — clean

Diff: 9 files, +26 / -33.

## Follow-up

Once this lands, PR #129 drops `continue-on-error: true` on the `mermaid-little Rust (fmt + clippy)` clippy step to make it gating.

Refs #123

### DIFF
--- a/crates/mermaid-little/src/parser/gantt.rs
+++ b/crates/mermaid-little/src/parser/gantt.rs
@@ -475,10 +475,9 @@ fn scan_str_after(s: &str, key: &str) -> Option<String> {
     let rest = rest.strip_prefix(':')?.trim_start();
     let (open, close) = if let Some(r) = rest.strip_prefix('"') {
         (r, '"')
-    } else if let Some(r) = rest.strip_prefix('\'') {
-        (r, '\'')
     } else {
-        return None;
+        let r = rest.strip_prefix('\'')?;
+        (r, '\'')
     };
     let end = open.find(close)?;
     Some(open[..end].to_string())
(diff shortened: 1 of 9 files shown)


kookyleo (approved):
已逐文件核对，通过。

**语义等价性**：9 处改写全部是 `else if let Some(x) = expr { .. } else { return None }` → `let x = expr?;` 的形式。两者在 `expr` 为 `None` 时都是提前 return `None`，在 `Some` 时都继续，控制流完全一致。

**`format!` 去 `&`**：`format_args!` 本身就按引用捕获参数，所以 `fill = task.fill`（`String` 字段）和 `color = l.node_colors[i]`（`Vec<String>` 索引）都不会发生 move，只是去掉了一层多余的显式借用。

**验证**：CI 全绿，其中 `mermaid-little Rust (fmt + clippy)` 与 `mermaid-little Rust tests` 均通过，正是这次要解锁的门禁本身。

PR #134, opened 2026-07-28
test(mermaid): split decluster zero-overlap contract from #93 repro

### LINKED ISSUE #126: [mermaid-little] Edge-label decluster follow-ups from the #122 review: overlap assertion, host wiring

mybee:
## Summary

Resolves #126 item 2 — the last open follow-up from the #122 review.

The strict zero-overlap assertion in the #93 repro integration test was brittle to layout drift: any dagre change that left a single pair overlapping turned the test red, and the fix instruction baked into the comment was a human judgement call each time (drift vs. regression). #128 added a weaker strict-decrease assertion *in front of* it but kept the zero-overlap assertion, so the brittleness remained.

This PR takes the "split the fixture" route from the review comment: the strict zero-overlap contract and the drift-robust strict-decrease property now live in tests that match what they actually prove.

- **Integration test on the #93 repro** (`decluster_reduces_edge_label_overlaps`, renamed from `..._removes_`) now asserts only strict-decrease — overlap count goes down. Robust to layout drift.
- **Strict zero-overlap contract** lives in the pass's unit test `result_has_no_remaining_overlaps`, which feeds `decluster` a synthetic rect cluster whose required displacement (~29 px) is far under `max_displacement` (80 px). `max_displacement` provably does not bind there, so a red zero-overlap result always means a regression, never drift.

Module doc comments on both the pass (`src/layout/edge_label_decluster.rs`) and the integration test updated to describe the split. No behaviour change — test + docs only.

## Why the unit test is the right home for the zero-overlap contract

The pass's own module doc notes it is "pure arithmetic over centre-based rects … trivially testable without laying out a real diagram." A real diagram cannot *provably* satisfy "max_displacement does not bind" — dagre's layout is empirical. Hand-constructed rects can: the cluster needs ≤29 px, the budget is 80 px, so the bound is an arithmetic fact, not a hope. That is exactly the "small synthetic case where max_displacement provably does not bind" the review asked for.

## Test plan

- [x] `cargo test -p mermaid-little --test edge_label_decluster` — 4 passed
- [x] `cargo test -p mermaid-little --lib edge_label_decluster` — 6 passed (incl. `result_has_no_remaining_overlaps`)
- [x] `cargo fmt -p mermaid-little -- --check` — clean
- [x] `cargo clippy -p mermaid-little --tests` — no new warnings in touched files (pre-existing `useless_borrows_in_formatting` in `svg_sankey.rs` / `svg_xychart.rs` untouched; clippy is non-blocking in CI)

Closes #126.

### DIFF
--- a/crates/mermaid-little/src/layout/edge_label_decluster.rs
+++ b/crates/mermaid-little/src/layout/edge_label_decluster.rs
@@ -18,10 +18,13 @@
 //! **Best-effort, not a guarantee.** The pass *reduces* overlap but does not
 //! promise zero overlaps: when `max_displacement` binds, labels that cannot
 //! be pushed far enough apart remain stacked. The property the pass actually
-//! provides is "overlap count strictly decreases"; "zero overlaps" is a
-//! fixture-specific fact that holds for some layouts (e.g. the #93 repro)
-//! and not others. Tests assert the general property, and only assert zero
-//! for a fixture known to fully resolve.
+//! provides is "overlap count strictly decreases"; "zero overlaps" holds
+//! only when `max_displacement` does not bind. Tests split the two: the
+//! integration test on the #93 repro asserts only strict-decrease (robust to
+//! layout drift), and the unit test `result_has_no_remaining_overlaps`
+//! asserts the strict zero-overlap contract on a synthetic cluster whose
+//! required displacement is far under `max_displacement` — so a red
+//! zero-overlap result always means a regression, never drift.
 //!
 //! The function is pure arithmetic over centre-based rects `(cx, cy, w, h)`
 //! so it is trivially testable without laying out a real diagram.
@@ -215,7 +218,16 @@ mod tests {
 
     #[test]
     fn result_has_no_remaining_overlaps() {
-        // A dense cluster of four labels plus a distant one.
+        // The strict zero-overlap contract. This feeds `decluster` a dense
+        // cluster of four 30x10 labels stacked within ~2 px of (50, 50) plus
+        // a distant outlier. Resolving the cluster needs at most
+        // ~(4 * 10 + 3 * gap) / 2 ≈ 29 px of displacement from the base
+        // centre — far under `max_displacement` (default 80 px), so
+        // `max_displacement` provably does not bind here and the pass is
+        // expected to fully resolve every overlap. A red assertion below is
+        // therefore always a regression, never layout drift. The
+        // integration test on the #93 repro asserts only strict-decrease,
+        // since that realistic fixture can leave `max_displacement` bound.
         let mut r = vec![
             (50.0, 50.0, 30.0, 10.0),
             (52.0, 52.0, 30.0, 10.0),
--- a/crates/mermaid-little/src/layout/edge_label_decluster.rs
+++ b/crates/mermaid-little/src/layout/edge_label_decluster.rs
@@ -220,11 +220,12 @@ mod tests {
     fn result_has_no_remaining_overlaps() {
         // The strict zero-overlap contract. This feeds `decluster` a dense
         // cluster of four 30x10 labels stacked within ~2 px of (50, 50) plus
-        // a distant outlier. Resolving the cluster needs at most
-        // ~(4 * 10 + 3 * gap) / 2 ≈ 29 px of displacement from the base
-        // centre — far under `max_displacement` (default 80 px), so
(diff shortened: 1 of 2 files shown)


kookyleo (approved):
已核对，通过。核心论点我在本地实测复核过。

**改动的实质**：把「零重叠」这个强断言从 #93 真实 repro 上摘下来，挪到合成用例上；真实 repro 只保留「重叠对数严格下降」。

**为什么这个拆分是对的**：`decluster` 是 best-effort 的 —— 当 `max_displacement` 触顶时，挤不开的 label 会保持重叠。所以「零重叠」在真实布局上是个会随布局漂移而失效的事实，不是算法契约；挂在那里迟早变成假红。

**实测复核**：我在本地给合成用例加了临时探针量实际位移，结果是

```
label 0: dy=-18.44   label 1: dy=+16.94
label 2: dy=-5.75    label 3: dy=+7.25
label 4: dy=0.00
最大位移 = 18.44 px / 预算 80.00 px
```

余量 4.3 倍，`max_displacement` 确实不可能绑定。所以注释里写的「~18 px」准确，「这里的零重叠结果本身就是预算未绑定的证明」这个推论也成立 —— 该用例一旦变红，只可能是算法回归，不会是漂移。

注释把两个测试各自守什么契约写清楚了，后来人不会再误读成通用保证。测试改名 `decluster_removes_...` → `decluster_reduces_...` 也和放宽后的断言对上了。

PR #141, opened 2026-07-28
chore: drop unused root pnpm-lock.yaml

kookyleo:
The root `pnpm-lock.yaml` is not consumed by anything, and had drifted far enough that it was actively misleading.

## Not used

The only pnpm step in CI is in `commonmark-conformance.yml`, and it sets `working-directory: tests/markdown-conformance`, installing against that directory's own `pnpm-lock.yaml` / `pnpm-workspace.yaml`. Every other job and script runs bun against `bun.lock`. A repo-wide search finds no other `pnpm install` at the root.

## Had gone stale

Last regenerated 2026-06-14, while `main` has moved on to 2026-07-27. In that window it fell behind in three ways:

- missing six workspace members added since — `crates/supramark-markdown/packages/{web,react-native}` and the four `packages/features/main/code-highlight*` packages
- still resolving `@kookyleo/graphviz-anywhere-rn@0.1.3` from the registry, where `bun.lock` links the in-tree `@actrium/*` workspace packages
- still aliasing `@kookyleo/graphviz-anywhere-web` as a link, likewise superseded

Anyone who ran `pnpm install` against it would have got a tree that does not match what CI builds.

## Also here

`REUSE.toml`'s lockfile annotation is widened from `pnpm-lock.yaml` to `**/pnpm-lock.yaml`, so the conformance harness's lockfile keeps its `CC0-1.0` annotation instead of falling through to the repository's Apache-2.0 default once the root pattern no longer matches anything.

`pnpm-workspace.yaml` is deliberately left in place — it still documents the workspace layout and mirrors `package.json#workspaces`.

## Verification

- `bun run license:check` — 64 manifests, all allow-listed
- no dangling references: the only remaining `pnpm-lock` mentions are the widened REUSE glob and `.prettierignore` (whose pattern still correctly covers the conformance lockfile)


boomsi (commented):
@kookyleo

Reviewed #141 hands-on in an isolated worktree (HEAD `794f145a`). Verdict: **ship it.**

**"Unused" claim — ✅ verified.** The only `pnpm install` in the repo is `commonmark-conformance.yml:88-90`, which sets `working-directory: tests/markdown-conformance` and installs against that dir's own `pnpm-lock.yaml`. Root `package.json` has no pnpm script; no dependabot/renovate config references the root lockfile. A repo-wide `find` confirms `tests/markdown-conformance/pnpm-lock.yaml` is now the only pnpm lockfile left, so widening the glob to `**/pnpm-lock.yaml` hits exactly that one file — no collateral.

**REUSE.toml widening — ✅ verified.** `reuse spdx` shows the conformance lockfile's `LicenseInfoInFile` going from `Apache-2.0` (it only matched the `**` default before — the literal `pnpm-lock.yaml` never matched a subdir path) to `CC0-1.0`, now consistent with `bun.lock` / `Cargo.lock`. `.prettierignore:17`'s `pnpm-lock.yaml` (slash-less → gitignore semantics) still matches the subdir file.

**Tests (ran manually):**
- `bun run license:check` → ✅ All 66 manifests allow-listed, exit 0.
- `reuse lint` → exit 1 both before and after, with a byte-identical SUMMARY (pre-existing deprecated `GPL-3.0`/`LGPL-3.0` + missing `CC0-1.0`/`EPL-2.0`/… license texts). `ci.yml:55` runs it `continue-on-error: true` (warn-only), so this PR neither introduces nor fixes a compliance change.
- PR CI: all jobs pass (License Compliance / Lint / Test & Build 20.x & 22.x / Rust suites / Graphviz release runtime ×3).

**Minor nits (non-blocking, doc-level only):**
- Body says "keeps its CC0-1.0 annotation" — empirically it's *granted*, not kept: the file was `Apache-2.0` before (via the `**` default, since the old literal never matched a subdir path). The "instead of falling through to Apache-2.0 default" half is accurate. Trivial framing fix if you care.
- The `64 manifests` figure in the Verification section is now `66`.

**One thing for maintainers' awareness (not a blocker, out of scope here):** root `pnpm-workspace.yaml` is kept while the root lockfile is gone. Anyone who runs `pnpm install` at the repo root will get a freshly-regenerated lockfile that won't match `bun.lock`'s in-tree `@actrium/*` aliases. Fine since pnpm-at-root isn't a supported path, but worth a conscious follow-up (drop the workspace file, or gitignore the root lockfile).

Ready to merge.

boomsi (approved):

PR #130, opened 2026-07-28
test: add cmark-gfm conformance suite and generalize workflow

alpacassarg:
## 修改内容

- 新增 cmark-gfm 官方规范及扩展用例，共 702 条
- 将用例转换为统一 JSON 格式，并在 `version.json` 中记录上游 commit
- 通用化用例导入、校验、语义测试、视觉测试和基线更新脚本
- 保留原有 CommonMark 命令作为兼容入口
- 报告内容根据数据源动态生成，并在 Markdown 中直接展示视觉差异图片
- 将 GitHub Actions 工作流改造成动态多数据源 matrix
- 支持选择运行 CommonMark、cmark-gfm 或全部数据源
- 各数据源独立上传报告并创建或更新对应的聚合 Issue
- 工作流更名为 `markdown-conformance.yml`

## 达成目标

建立了统一、可扩展的 Markdown 规范测试框架。后续增加新的规范用例时，只需添加数据源配置、导入器和统一用例文件，无需再复制整套测试及工作流逻辑。

PR #143, opened 2026-07-29
fix(parser): match emphasis in content coordinates

kookyleo:
Found while looking into the one red check on #130. The `cmark-gfm 语义与视觉对照` job reports `Rust parser did not produce an AST` for the "Embedded pipes" case — that is not a harness problem, the parser panics.

## The bug

```
thread 'main' panicked at crates/supramark-markdown/src/generics/inline/emph_pair.rs:143:
attempt to subtract with overflow
```

Three lines are enough:

```markdown
| a |
| - |
| _x\|y_ |
```

The table cell scanner hands the inline parser the **unescaped** cell content, so `state.src` holds `_x|y_` (5 bytes) for the source text `_x\|y_` (6 bytes), with a `(content_pos, source_pos)` mapping table bridging the two. `InlineState::pos` indexes the content; `Node::srcmap` holds source offsets. They are different coordinate spaces and the sizes do not agree.

`EmphPairScanner::run` mixed them:

```rust
let map = node.srcmap.unwrap().get_byte_offsets();  // source coordinates
state.pos += scanned.length;                        // content coordinates
let token_len = map.1 - map.0;                      // a source-space length
state.pos -= token_len;                             // 5 - 6
```

The escape is the only thing needed to make the two spaces disagree — `| x\|y |` alone is fine, and `~x\|y~` is fine, because only this rule reads a length back out of a srcmap.

Severity is not limited to the test harness. `usize` subtraction overflow only panics in debug builds; a release build wraps to a value near `usize::MAX` and carries on with that as `state.pos`, so the napi and wasm paths get either a panic on the next slice or silently wrong output. cmark-gfm's own test suite is what surfaced it, but nothing about the input is exotic — an escaped pipe inside emphasis inside a table cell is ordinary GFM.

## The fix

Delimiter matching now stays in content coordinates from start to finish, and crosses into source coordinates exactly once, at the end, through `get_map`.

- `EmphMarker` carries `content_start` / `content_end` for the still-unmatched part of its delimiter run. `Node::srcmap` cannot be reused for this, which is the whole point.
- `scan_and_match_delimiters` cuts `marker_len` off those offsets rather than off srcmaps, and returns the resulting token's content start. `run` no longer reads a length back out of a srcmap, so there is nothing left to underflow.
- `InlineState::get_pos_for_source` inverts the existing `get_source_pos_for`, for callers that genuinely need to cross between the spaces.

This also corrects the source maps of the emph nodes created inside the matching loop. They were being mapped twice — `state.get_map` was called with offsets that were already in source coordinates. That was harmless in practice only because the old code used their *difference*, where a constant per-line offset cancels out; the absolute values were wrong for any emphasis not at offset 0. Those maps are now single-mapped and correct.

## Verification

Everything below was run, not reasoned about.

**No behaviour drift on CommonMark.** Failure sets compared case-by-case with and without the patch on the same tree:

| Path | Before | After |
|---|---|---|
| `run-commonmark.mjs` (ast-projection) | 640/652, 12 failures | 640/652, **same 12 ids** — `newly broken: []`, `newly fixed: []` |
| `run-commonmark-visual.mjs` (production-web-renderer-dom) | — | **652/652 semantic, 652/652 visual** |

The visual numbers match what CI reports green on `main`, which is the target `baselines/commonmark.json` is keyed to.

**cmark-gfm goes from 1 execution error to 0.** Run against #130's fixtures, same comparison target CI uses:

| | #130 CI, no patch | local, with patch |
|---|---|---|
| semantic | 644 passed / 57 failed / **1 error** | 644 passed / 58 failed / **0 errors** |
| visual | 672 passed / 29 failed / **1 error** | **673 passed** / 29 failed / **0 errors** |

The case stops crashing and becomes an ordinary semantic mismatch; visually it now actually renders correctly, hence 672 → 673.

**Source maps are right, not just non-crashing.** For the minimal case, `emphasis` maps to `[14,20)` = `_x\|y_` and the inner `text` (`x|y`) maps to `[15,19)` = `x\|y` — i.e. onto the escaped source, which is the point of the mapping table.

**Suite:** `cargo test -p supramark-markdown` — 59 + 27 passing, including a new `public_api` regression test that asserts both the position mapping and the unescaped text value. `cargo fmt --check` clean. `cargo clippy -D warnings` leaves 4 findings in `full_link.rs`, `list.rs` and `reference.rs` that are present on `main` unchanged and untouched by this patch.

## Why this matters for #130

#130 ships no `baselines/cmark-gfm.json`, so its report says `baseline: { configured: false, reason: "baseline-missing" }`. With `FAIL_WORKFLOW` defaulting to true, merging it makes the conformance workflow red on every push to `main` — as a flat "58 not passing", not as "regressed against a baseline".

The normal way out is to record a baseline and ratchet from there, but `update-baseline.mjs` refuses:

```js
if (!summary.visual?.enabled || summary.errors !== 0 || summary.visual.errors !== 0) {
  throw new Error('拒绝更新异常运行基线：必须完成视觉测试且语义、视觉执行错误均为 0。');
}
```

`errors` was exactly 1, and it was this panic. With it fixed both counts are 0 and a baseline can be recorded, so #130 can land green with a working ratchet instead of a permanently red workflow. Landing this first is the cheaper order; nothing here depends on #130.

### DIFF
--- a/crates/supramark-markdown/src/generics/inline/emph_pair.rs
+++ b/crates/supramark-markdown/src/generics/inline/emph_pair.rs
@@ -82,6 +82,17 @@ pub struct EmphMarker {
     // an emphasis.
     pub open: bool,
     pub close: bool,
+
+    // Byte range of the still-unmatched part of this delimiter run, in
+    // `InlineState::src` coordinates.
+    //
+    // `Node::srcmap` cannot be reused for this: it holds offsets into the *original* markdown
+    // source, which is a different (and possibly longer) coordinate space. A table cell hands
+    // the inline parser `_x|y_` for the source text `_x\|y_`, because the cell scanner drops
+    // the backslash. Matching delimiters has to be done in `src` coordinates, and only the
+    // final span is translated to source coordinates via `InlineState::get_map`.
+    pub content_start: usize,
+    pub content_end: usize,
 }
 
 // this node is supposed to be replaced by actual emph or text node
@@ -127,19 +138,27 @@ impl<const MARKER: char, const CAN_SPLIT_WORD: bool> InlineRule
         }
 
         let scanned = state.scan_delims(state.pos, CAN_SPLIT_WORD);
+        let content_end = state.pos + scanned.length;
         let mut node = Node::new(EmphMarker {
             marker: MARKER,
             length: scanned.length,
             remaining: scanned.length,
             open: scanned.can_open,
             close: scanned.can_close,
+            content_start: state.pos,
+            content_end,
         });
-        node.srcmap = state.get_map(state.pos, state.pos + scanned.length);
-        node = scan_and_match_delimiters::<MARKER>(state, node);
-        let map = node.srcmap.unwrap().get_byte_offsets();
-        // backtrack to keep correct source maps
-        state.pos += scanned.length;
-        let token_len = map.1 - map.0;
+        node.srcmap = state.get_map(state.pos, content_end);
+        let token_start;
+        (node, token_start) = scan_and_match_delimiters::<MARKER>(state, node);
+
+        // Backtrack to keep correct source maps: `InlineParser::tokenize` overwrites the srcmap
+        // of whatever we return with `get_map(state.pos - len, state.pos)`. So `len` has to be
+        // the token's length in `state.src` coordinates, and `state.pos` has to end up just
+        // past the closing marker run.
+        state.pos = content_end;
+        debug_assert!(token_start <= state.pos);
+        let token_len = state.pos - token_start;
         state.pos -= token_len;
         Some((node, token_len))
     }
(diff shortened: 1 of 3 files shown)


mybee (approved):

boomsi (commented):
@kookyleo

**Re-review: core fix verified hands-on — ready to merge, two non-blocking cleanups**

Verified at HEAD `21256557` (isolated worktree, everything run locally, not just CI):

- ✅ `cargo test -p supramark-markdown` — 59 unit + 27 public_api, all pass (incl. new `public_api_maps_emphasis_spanning_an_escape_inside_a_table_cell`)
- ✅ CommonMark semantic — 640/652, 12 failures, **0 drift vs main** (same 12 IDs: 0009/0126/0130/0144/0237/0294/0296/0307/0319/0320/0321/0323; newly broken/fixed both empty); emphasis section 132/132
- ✅ CommonMark visual — 652/652 semantic + 652/652 visual (ran `run-commonmark-visual.mjs` with Chromium installed)
- ✅ cmark-gfm "Embedded pipes" panic fixed — execution errors **1 → 0**; the case stops crashing and becomes an ordinary semantic mismatch (ran with #130's conformance applied + PR 143 parser). Minimal case `_x\|y_` maps emphasis→[14,20) and inner text→[15,19) value=`x|y` onto the escaped source, exactly as described
- ✅ release build — `cargo build --release` bin runs the minimal case with no panic/underflow; AST matches debug
- ✅ native FFI + wasm — `cargo build -p supramark-markdown-native` and `bun run build:wasm markdown` both succeed (PR doesn't touch binding code or `parse()` signature)
- ✅ `cargo fmt --check` clean

🟡 Please consider before merge (non-blocking):

1. `InlineState::get_pos_for_source` (`state.rs:256`) ships as public API with **zero call sites** across the crates (its inverse `get_source_pos_for` has 4). It's correct — saturating guard, monotonic — but outside this PR's scope, since the emph fix stays in content coordinates end-to-end and never needs to cross back. Please clarify intent or split it into its own change.
2. The new test introduces 5 `redundant_reference_in_panic_argument` clippy warnings at `public_api.rs:681/684/687/693/704` (`&children[..]` passed to `{:?}`). Existing tests use inline `{var:?}` to avoid this — trivial to match. Note this diverges from the PR's "clippy leaves only 4 pre-existing findings": those 4 are confirmed pre-existing in `full_link.rs`/`list.rs`/`reference.rs`, and the PR-touched files (`emph_pair.rs`, `state.rs`) are clean.

Tests:
- `cargo test -p supramark-markdown`: 59 unit + 27 public_api = 86 passed, 0 failed
- CommonMark semantic: 640/652 (12 fail, 0 error), 0 drift vs main
- CommonMark visual: 652/652
- cmark-gfm semantic: 631/702, **0 execution errors** (was 1 on main; "Embedded pipes" no longer panics)
- `cargo fmt --check`: clean; `cargo clippy --all-targets`: 4 pre-existing lib + 5 new test warnings

Ready to merge. The two cleanups are non-blocking — fine as-is or addressed first. cmark-gfm panic fix independently reproduced (errors 1→0). Notes: cmark-gfm visual (672→673) was not re-run locally (semantic already proves the panic fix); the absolute cmark-gfm pass count 631 vs the PR's 644 is a #130 conformance version/config difference, not a PR 143 effect — main and PR 143 are both 631, differing only in the Embedded pipes error→diff.

Issue #139, opened 2026-07-28
[d2-little] Ship a browser playground for the wasm build


kookyleo:
> **Migrated from [kookyleo/d2-little#4](https://github.com/kookyleo/d2-little/issues/4)**, originally requested by **@Mapaor** on 2026-05-26. GitHub cannot transfer issues across owners (user → org), so it was re-filed here by hand. d2-little has been subtree-merged into this repository at `crates/d2-little/`.

## Original request

> Looks nice. Given you already can compile to wasm32, it would be great if you created a simple website that serves as a d2-little playground to test it out.

## Context in this repository

The wasm build the request refers to is `crates/d2-little/packages/web` (`@actrium/d2-little-web`), and there is already a browser sandbox at `examples/react-web-csr` that consumes it through `@supramark/engines`, plus `bun run feature:preview:web`. So the pieces exist; what is missing is a public, deployed page where someone can paste D2 source and see SVG without cloning the repo.

Worth deciding as part of this:

- whether the playground covers d2 only, or all the `*-little` engines (d2 / mermaid / plantuml / graphviz) behind one page — the engine interface in `packages/engines` is already uniform, so a shared playground is barely more work than a d2-only one
- where it gets hosted, and whether the wasm artifacts are published to npm first (currently `@actrium/*` wasm packages are unpublished — see #137)

- [ ] decide single-engine vs unified playground
- [ ] decide hosting
- [ ] build and deploy

Mapaor:
okay but you should link the demo in the main readme somewhere

PR #145, opened 2026-07-30
feat(d2-playground): ship browser playground for d2 wasm

mybee:
## Summary

Refs #139. (Keeping the issue open: its "single-engine vs unified playground" decision is still unresolved, and this PR ships the d2-only variant.) Adds a public, deployable D2 playground: paste D2 source, see the SVG — no repo clone required.

- **`examples/d2-playground/`** — d2-only Vite app (left editor + right SVG preview, example snippets incl. `layout-engine: elk`, localStorage persistence, SVG download). Calls `loadWebD2Render()` directly from `@supramark/engines/web` instead of the full `createWebDiagramEngine` facade, so plantuml/mermaid/graphviz wasm never ships with the playground.
- **Vite config** aliases the three unused `@actrium/*` engine packages to an empty stub. Rollup resolves their dynamic imports during the module-graph build (before tree-shaking); the stub satisfies resolution so the build doesn't require those wasm dists. The d2 render path only touches `@actrium/d2-little-web` + `elkjs`.
- **Deploy** via the docs site rather than a standalone Pages workflow — the Pages root is already owned by `docs.yml`, and a second `deploy-pages` workflow would overwrite it. New `docs:playground:build` script builds the app into `docs/public/playground` with `base /supramark/playground/`, chained into `docs:preview:build`. `docs.yml` path trigger covers `examples/d2-playground/**`.
- Root `workspaces` gains `examples/d2-playground`; `docs/public/playground` is gitignored (build artifact, regenerated by CI).

URL once merged: https://actrium.github.io/supramark/playground/

## Test plan

- [x] `bun install` + `bun run build:wasm` (d2-little-web wasm built)
- [x] `cd examples/d2-playground && bun run build` — passes; only d2 wasm + elk.bundled emitted, no plantuml/mermaid/graphviz wasm leak
- [x] `bun run dev` boots clean; all modules transform (HTTP 200)
- [x] Headless Chrome DOM dump confirms a real d2 SVG renders in the preview pane (`<svg ... data-d2-version="v0.7.1-HEAD">`)
- [x] `bun run docs:playground:build` — builds into `docs/public/playground/` with correct `/supramark/playground/` asset base
- [ ] After merge: docs.yml deploys; visit https://actrium.github.io/supramark/playground/ and verify the four example snippets render (basic flow, shapes, elk layout, containers) and an invalid input shows the error pane


kookyleo (approved):
## Review: approving — builds green, no new dependencies, no CJK, and the wasm isolation trick actually works

Branch refreshed against current main before merging, so this now has a real CI run against the gates it originally predated (English-only source, cmark-gfm conformance) rather than a green from 19 commits ago.

**Build verified, not assumed.** `vite build --base /supramark/playground/` in `examples/d2-playground`: exit 0, 1919 modules, 45.29 s. Emitted `d2_little_web_bg.wasm` 2788 kB, `elk.bundled.js` 2276 kB, `index.js` 314 kB. `ls dist/assets | grep -Ei 'plantuml|mermaid|viz|graphviz'` returns nothing — the empty-stub alias in `vite.config.ts:36-47` genuinely keeps the other three engines' wasm out of the bundle, which was the claim most worth checking. Built `index.html` emits `/supramark/playground/assets/...`, so the base path is right. ESLint clean at `--max-warnings 0`.

**English-only gate: clean.** Every added file under `examples/d2-playground/` is ASCII, including UI strings, comments, CSS and the four D2 snippets. The one CJK hit across the whole diff is `package.json:5`'s `description`, which is byte-identical on main and untouched here. Running main's `check-cjk.mjs` filtered to this PR's files: 0 findings. (Separately: that root `description` slips the gate only because `SOURCE_EXTENSIONS` in `scripts/check-cjk.mjs:76` excludes `.json`, yet it is an npm-facing string. Not this PR's problem, but worth its own fix.)

**Zero new third-party dependencies.** The `bun.lock` diff adds the workspace entry plus two nested `@types/react*` resolutions; `vite`, `@vitejs/plugin-react`, `vite-plugin-wasm`, `vite-plugin-top-level-await` and `typescript` all already existed at top level, declared identically by `examples/react-web-csr`. `license-check` passes — `examples/d2-playground/package.json:4` declares `Apache-2.0`.

**The docs workflow wiring is correctly ordered.** `docs.yml:52-59` installs the Rust toolchain with `wasm32-unknown-unknown` and `wasm-pack`; `docs.yml:65` runs `docs:build` -> `docs:preview:build`, whose first command is `bun run build:wasm`; `scripts/build-wasm.ts:38-42` includes the d2 target. The artifact the playground needs is built before it is needed.

**No bare-specifier leak** of the `#79`/`#80` class: the two surviving `@actrium/d2-little-web` strings in the bundle are an error message and a JSX label, not module specifiers. Error handling is sound — the loader rejection is caught, the render is debounced at 250 ms with a proper `cancelled` guard, and because `crates/d2-little/packages/web/src/lib.rs:33` returns `Result<String, JsValue>`, D2 parse errors arrive as a thrown JS *string*, correctly handled by `e instanceof Error ? e.message : String(e)`.

### Scope note

This ships the d2-only variant while #139's first checklist item — single-engine vs unified playground — was still open, and #139's own text leans toward unified on the grounds that the `packages/engines` interface is already uniform. I have changed the description from `Closes #139` to `Refs #139` so the issue stays open to track the unified version; the hosting decision in the description is sound and I verified the reasoning (`docs.yml` already owns `concurrency: group: pages`, so a second Pages workflow really would contend). Note that the stub-alias design is precisely what a unified playground would have to unwind.

### Follow-ups (non-blocking)

- **`package.json:50` puts the playground build on the critical path of the entire docs deployment.** `docs:preview:build` now ends with `docs:playground:build`, and `docs:build` chains after it, so a failing `vite build` here means typedoc and vitepress never run and nothing deploys. `react-web-csr` already has this coupling, so it is not a new class of risk, but it doubles the exposure. A separate step, or a non-fatal one, would be safer.
- **Naming:** `docs:playground:build` is nested inside `docs:preview:build`, but "preview" is the `react-web-csr` output at `/supramark/preview/` while this lands at `/supramark/playground/`. It belongs in `docs:build` alongside `docs:typedoc`.
- **`.eslintrc.js:90-97` was not updated** — `docs/public/preview/` and `docs/public/typedoc/` are in `ignorePatterns` but `docs/public/playground/` is not, though `.gitignore:66` did get the entry. Anyone who runs `docs:build` then `lint` locally gets ESLint parsing the 313 kB and 2.28 MB minified chunks under `--max-warnings 0`. CI is unaffected because the lint job never builds docs.
- **Nothing type-checks this app.** `"build": "vite build"` with no `tsc`, and the shipped tsconfigs do not actually pass: `tsc -p tsconfig.app.json --noEmit` exits 2 with `TS1294` at `packages/engines/src/engine.ts:22` (a parameter property, rejected by `erasableSyntaxOnly: true`). `scripts/lint-types.ts` only walks `packages` and `crates`. `react-web-csr` has the identical config and the identical omission, so this is house style faithfully copied rather than a regression — but the type-safety guarantee for both example apps is zero.
- **`App.tsx:64` holds a `DiagramRenderFn` in a `useRef<RenderFn | null>`.** `RenderOptions` has no index signature so it is not assignable to `Record<string, unknown>`; this compiles only because `strict: false` disables `strictFunctionTypes`. Importing `DiagramRenderFn` instead is free.
- **`examples/d2-playground/package.json` does not declare `elkjs`** although the "ELK layout" example needs it; it resolves only by hoisting out of `@supramark/engines`' devDependencies. Works today, would break under a stricter or isolated install. Same undeclared reliance exists in `react-web-csr`.
- **`App.tsx:193`** shows "Loading engine…" whenever `svg` is empty and `error` is null, including after the engine is ready and the user clears the textarea. It should key off `engineReady`.
- **No nav entry** in `docs/.vitepress/config.mts`, so the page is reachable only by typing the URL. The existing preview app is unlinked too, so this is consistent — but it undercuts #139's goal of a public page someone can just open.
- `App.tsx:190` uses `dangerouslySetInnerHTML` on engine SVG output. `loadInitial()` reads only `localStorage`, never a URL query or hash, so there is no reflected-input vector — self-XSS at worst, and the same pattern as `DiagramBlock.tsx:48`. The one thing that makes it non-zero is that the deploy target shares an origin with every other Pages project in the org. Parse-and-reserialize, or DOMPurify with `USE_PROFILES: {svg: true}`, if you want it closed.
- `bun.lock:3` adds `"configVersion": 0`, absent from main's lock — written by a newer Bun than the 1.3.14 pinned in CI. Harmless (CI does not use `--frozen-lockfile`) but the lock will churn.

> Verification scope: no browser was launched. The four example snippets rendering, and mid-keystroke invalid D2 degrading gracefully rather than poisoning the wasm instance, remain unverified — which matches this PR's own unchecked test-plan box. Note there is no `console_error_panic_hook` registered, so a Rust panic would surface opaquely; that is a d2-little concern rather than this PR's.

PR #148, opened 2026-07-30
chore: make the repository English-only and enforce it in CI

kookyleo:
Makes the repository English-only and enforces it in CI, so a future non-English commit is stopped rather than found later.

Four commits, meant to be read in order:

| | |
|---|---|
| `ci:` | adds `scripts/check-cjk.mjs` and wires it into the Lint job |
| `refactor:` | 192 source files translated to English |
| `docs:` | regenerated generated docs |
| `docs:` | 51 Chinese documents renamed to `.zh.md`, plus an English root README |

## The gate

CJK is allowed in `*.zh.md`, in `i18n/ locales/ translations/`, in `tests/fixtures/ testdata/ tests/cases/_fixtures/`, in committed build output, and in untracked files. Anywhere else fails the build. Non-source extensions (`.svg`, `.mmd`, `.d2`, `.json`) are left alone, since CJK there is subject matter rather than prose a reviewer has to read.

Where CJK is functionally required, an inline pragma exempts it and **the reason is mandatory** — a bare `cjk-allow:` is rejected, because the reason is the part a reviewer needs in order to judge the exemption:

```
// cjk-allow: <reason>        that line and the next
// cjk-allow-file: <reason>   the file, declared in the first 40 lines
```

**It runs inside the Lint job on purpose.** Lint is already a required status check on `main`, so the gate blocks a merge without anyone having to edit the branch ruleset. It also runs under `if: always()`, so a language violation is still reported when eslint is what failed, and it is a named step rather than a generic lint failure.

Pragma handling was checked against seven cases, including the two that must **not** be exempt: a pragma with no reason, and a file pragma declared past the scan window.

## What keeps its Chinese, and why

Eight files carry a justified exemption. These are the ones where translating would change behaviour, not just wording:

- `crates/plantuml-little/src/text.rs` — `display_width` unit tests
- `crates/mermaid-little/tests/edge_label_decluster.rs` — CJK edge labels for #93
- `packages/core/__tests__/fixtures/cjk-samples.ts` — a multi-byte sample lifted out of `parser.test.ts` so the exact bytes stay asserted
- `tests/official-diagram-visual/cases/official-diagram-rendering-cases.md` and its v2 sibling — **parsed as data.** The workflow loads them by hard-coded filename and matches their Chinese section headers with regexes. Verified after the change that both still yield 30 cases and 10 category headings, byte-order marks intact.
- the two `build-official-diagram-cases*.mjs` generators, which must keep emitting headers that parser matches
- `CLAUDE.md`, `AGENTS.md` and the directory-scoped `AGENTS.md` — looked up by exact filename by the tooling that reads them, so `.zh.md` is not an option

## Deliberate output changes

- **The Markdown conformance harness reports in English.** `summary.result` moves from the Chinese pass/fail words to `'pass'`/`'fail'`, and every site comparing it was updated in the same change. Report metadata locale follows to `en-US`. The Playwright browser locale stays `zh-CN`, because it affects rendered pixels and the visual baselines are keyed to it.
- **The official-diagram-visual harness reports in English, including issue titles.** It finds an existing issue by exact title equality with no marker fallback, so #98 will no longer be matched and the next run will file an English duplicate. Worth closing #98 when this lands. The conformance harness does not have this problem — it matches on a marker, which was left untouched.
- Conformance workflow and job names are English. Neither is a required status check, so no ruleset change is needed.
- The VitePress site now has English navigation while still declaring `lang: 'zh-CN'`. Left alone deliberately: that is a locale decision, not a CJK removal.

## The root README

`README.md` is English and `README.zh.md` holds the original verbatim, with links between them. Renaming alone would have left GitHub showing no README at all.

It is **not** a literal translation, because the Chinese one had drifted from the code — it described `packages/rn` rather than `packages/renderers/rn`, an `examples/react-web` that does not exist, a flat `packages/features/*/feature-*` layout that is now three levels deep, and two documentation links that were already dead. The English version describes the tree as it is, and every link in it resolves. The original's stale claims are preserved in `README.zh.md` rather than silently corrected there.

## Fixes the rename and translation forced

Each of these would have broken something quietly:

- **`feature-update.ts`** detected legacy features by matching Chinese guidance text in the very `feature.ts` files this change translates, so the check would have matched nothing. It now matches the English wording. It never matched generator output before either, so its behaviour is unchanged.
- **The feature linters required `README.md` by exact name**, so a translated README counted as missing for 15 packages. Both now accept either name.
- **`resolve-sources.mjs` kept splitting the source list on a fullwidth comma**, written as an escape. Dropping it would have broken a list typed with a CJK IME.
- **`check-cjk.mjs` skips committed build output.** Its comments are a transpiled copy of source that is already checked, and it cannot be hand-fixed safely because edits shift the line numbers its `.js.map` is keyed to. The bob target directories are matched specifically rather than a bare `lib/`, since `tests/markdown-conformance/lib/` is hand-written source.
- **`check-cjk.mjs` no longer flags the language switcher** in the English half of a translated pair — a link whose target is the `.zh.md` counterpart. Only that link is discounted, so Chinese prose in an English document still fails; both directions are covered by a check. Without this, the bilingual `mermaid-little` and `plantuml-little` READMEs would have been "fixed" by overwriting their real translations.

Reference updates that came with the renames: markdown links, VitePress nav and sidebar routes, 15 `package.json` `files` arrays, `Cargo.toml` keys, the `REUSE.toml` MPL-2.0 annotation path, `typedoc.json`, and the getting-started link baked into the doc generators.

## Verification

| | |
|---|---|
| `check-cjk` | clean, 0 findings |
| `cargo test -p supramark-markdown` | 27 passed |
| engines / rn / web | 20 / 90 / 25 pass, 0 fail |
| `features:lint` | 21/21 |
| `license:check` | 66 manifests |
| VitePress build | 0 errors |
| diagram-visual self-test | 11 checks |
| conformance runner | still produces a report and a gate verdict |

`bun run lint` fails locally on three eslint errors in **untracked** vendored graphviz build output under `crates/graphviz-anywhere/graphviz/windows/…`. Those files are in no commit, so CI never sees them; eslint reports no errors across tracked files.

## Pre-existing defects found, deliberately not fixed here

Flagged rather than folded in, to keep this change reviewable:

- `crates/plantuml-little/src/text.rs` is never declared as a module anywhere in that crate, so it is dead code and its `display_width` tests exercise nothing.
- `feature:lint <name>` requires `jest.config.cjs`, which 17 of 21 feature packages do not have — so it already failed for most packages before this change.
- Around nine dead documentation links, including two in the old root README and a cluster in `CREATE_FEATURE_GUIDE` / `FEATURE_QUALITY_ASSURANCE` pointing at files that exist nowhere in the repo.
- `project-quality.ts` calls `process.chdir()` in a loop but only restores the working directory once at the end of `main()`, so a throw mid-loop leaves later iterations on the wrong directory.
- `feature-del.ts`'s `removeFromBundles()` builds regexes with double-escaped `\\n` and `\\s` inside real regex literals, so it likely never matches what it intends.
- `bun run features:sync` recreates `examples/*/src/all-features.ts`, which were deliberately deleted and are not gitignored, so the command leaves untracked files behind.
- The #106 scale-only size-delta exemption still classifies a diagram shrunk to a few percent of its area as a pass, and the self-test still encodes that as intended.
- `packages/renderers/rn-selection/…/selectable-rich-text` cannot be rebuilt: `bob build` fails on a pre-existing type error at `src/SelectableRichText.tsx:180` against React Native's generated `children` type.

## Note on scope

The translation ran across ten parallel agents. Every batch was verified to zero findings and its own tests re-run, and the cross-batch couplings — the `feature-update.ts` detection, the web/rn diagram status strings and their asserting test, the case-parser trio — were reconciled and re-verified afterwards rather than assumed.


boomsi (commented):
@kookyleo

## Re-review: verified — looks good to merge

Verified in an isolated worktree at HEAD `3a2eb54a`:

- ✅ `node scripts/check-cjk.mjs` → clean, 0 findings, exit 0 (8 file-pragma exemptions, matching the 8 justified files in the PR body)
- ✅ `cargo test -p supramark-markdown` → 27 passed, 0 failed
- ✅ `bun --filter @supramark/engines test` → 20 pass, 0 fail
- ✅ `bun --filter @supramark/rn test` → 90 pass, 0 fail
- ✅ `bun --filter @supramark/web test` → 25 pass, 0 fail
- ✅ `bun run features:lint` → 21/21
- ✅ `bun run test:core` → 89 pass, 0 fail
- ✅ `bun run license:check` → 66 manifests
- ✅ `vitepress build docs` → build complete, 0 errors
- ✅ case-parser trio three-way alignment confirmed by node decode — the workflow regex, the generator emit, and the `cases/*.md` literal all decode to `官方来源：` / `官方渲染 URL：` / `建议检查文本：` (`official-diagram-visual-workflow.mjs:268-270`, `build-official-diagram-cases.mjs:500-516`)
- ✅ conformance `summary.result` self-consistent — assignment uses `'pass'`/`'fail'`, `run.mjs` compares `=== 'fail'`, and no residual Chinese pass/fail comparisons remain anywhere in the suite
- ✅ `feature-update.ts` `multiNodeTypeGuidance` behaviour unchanged — neither the old Chinese nor the new English phrases exist in the template or any `feature.ts`, so the check stays `false` either way (a pre-existing dead check, not introduced here)

Not verifiable locally (need Playwright / heavy toolchain) — CI covers these:
- 🟡 diagram-visual self-test (11 checks) — `official-diagram-visual-workflow.mjs:9` loads chromium at import time; no playwright installed locally
- 🟡 conformance runner report + gate verdict — same Playwright dependency
- 🟡 full `bun run docs:build` chain — only the final `vitepress build docs` step was run locally

Heads-up you already flagged in the PR body, restating so it's tracked: #98 will no longer match after the English issue-title change, so the next run will file an English duplicate — worth closing #98 when this lands.

Tests: cargo 27/27 · core 89/89 · engines 20/20 · rn 90/90 · web 25/25 · features:lint 21/21 · license 66 · check-cjk clean · vitepress 0 errors.

Ready to merge.

boomsi (approved):

PR #149, opened 2026-07-30
chore(mermaid-little): bump 4 transitive deps in tests/support (#140 Tier 1)

### LINKED ISSUE #140: [mermaid-little] tests/support carries the same outdated npm deps a dependabot PR flagged upstream
> **Migrated from [kookyleo/mermaid-little#1](https://github.com/kookyleo/mermaid-little/pull/1)** (dependabot, 2026-07-24). The upstream repo is being retired in favour of this one, so the PR cannot simply be merged there — but the finding applies here unchanged.

`crates/mermaid-little/tests/support/` has the same locked versions the upstream PR wanted to bump. Verified against `package-lock.json` in this repo:

| Package | Locked here | Proposed upstream |
|---|---|---|
| `mermaid` | 11.14.0 | 11.15.0 |
| `dompurify` | 3.4.1 | 3.4.12 |
| `form-data` | 4.0.5 | 4.0.6 |
| `uuid` | 11.1.0 | 14.0.1 |
| `ws` | 8.20.0 | 8.21.1 |

## Why this is not a blind bump

`tests/support` is the deterministic Node-based reference SVG generator (`generate_ref.mjs`). mermaid-little's test suite compares its Rust output **byte-exactly** against SVGs produced by this generator, so **bumping `mermaid` 11.14.0 → 11.15.0 changes the baselines** — 11.15.0 includes rendering changes (e.g. decimal `autonumber` in sequence diagrams). That bump means regenerating and reviewing the reference set, not just editing a version string.

The other four are transitive dependencies of the generator and do not affect its output.

## Risk framing

The package is `private: true`, is never published, and only runs locally/in CI to regenerate reference SVGs — it is not on any consumer's dependency path. So this is hygiene rather than an exposure in shipped code. It still deserves a decision rather than being dropped on the floor when the upstream repo is archived.

- [ ] bump `dompurify`, `form-data`, `uuid`, `ws` (no baseline impact expected — confirm by re-running the reference suite)
- [ ] decide separately whether to move to `mermaid` 11.15.0 and re-baseline, or stay pinned on 11.14.0 and document why
- [ ] apply the same check to `crates/plantuml-little/tests/support/` and `crates/d2-little` test helpers

mybee:
## Summary

Resolves #140 (migrated from `kookyleo/mermaid-little#1`) — Tier 1 deps, the Tier 2 mermaid decision, and the cross-crate audit all land here.

### Tier 1 — 4 transitive dep bumps

Bumps the 4 transitive dependencies of the private `tests/support/` reference SVG generator that the upstream dependabot PR flagged, via an `overrides` block in `crates/mermaid-little/tests/support/package.json`:

| Package | Before | After | Pulled in by |
|---|---|---|---|
| `dompurify` | 3.4.1 | 3.4.12 | `mermaid` `^3.3.1` |
| `form-data`  | 4.0.5 | 4.0.6   | `jsdom` `^4.0.0` |
| `ws`         | 8.20.0 | 8.21.1 | `jsdom` `^8.18.0` |
| `uuid`       | 11.1.0 | 14.0.1 | `mermaid` `^11.1.0` (major bump, forced via override) |

All four are transitive, so an `overrides` field is required to force the requested versions (a bare `npm install pkg@ver` would promote them to direct deps; `uuid` 14 also exceeds mermaid's `^11` range).

`tests/support` is `private: true`, never published, only runs locally/in CI to regenerate reference SVGs — this is hygiene, not a shipped exposure.

### Tier 2 — mermaid stays pinned at 11.14.0 (decision recorded)

`mermaid` 11.14.0 → 11.15.0 is **deliberately not bumped**. A bump is not a version-string edit here: it regenerates and re-reviews the entire reference set, and 11.15.0 additionally breaks the generator on two fronts:

- **Baseline drift** — 11.15.0 includes rendering changes (e.g. decimal `autonumber` in sequence diagrams); every baseline SVG would shift and need re-approval.
- **jsdom generator breakage** — 11.15.0's stylesheet path hits the read-only `CSSStyleSheet` getter in jsdom 25.0.1, which the generator's shims don't cover. Fixing it means widening the shims, which diverges the generator's DOM environment from the Rust-side `stylis.rs` CSS parity surface the reference set validates.

So the call is: stay pinned on 11.14.0 (the version `tests/reference/VERSION` anchors the byte-exact comparison to) and revisit only when someone is prepared to re-baseline and re-shim together. This rationale is recorded in `crates/mermaid-little/tests/support/README.md` under "Upstream dep bumps (issue #140)".

### Cross-crate audit — clean

The same five-package check was applied to the sibling crates:

- `crates/plantuml-little/tests/support/` — `npm ls dompurify form-data uuid ws mermaid` is empty; the dep tree carries none of the five (only `@kookyleo/graphviz-anywhere-web`). No action.
- `crates/d2-little/` — no support `package.json` with dependencies; MathJax is vendored (`mathjax.js`), `elkjs` is resolved from the workspace `engines` package. No action.

## Verification — zero baseline impact

The whole point of Tier 1 is that these bumps must not change the byte-exact reference SVGs. Proven by control:

1. Regenerated the full 1394-fixture reference tree (`node generate_ref.mjs --batch` under Node 20) **with the overrides** → diff vs committed baselines: `904 files changed, 1039 insertions(+), 1039 deletions(-)`.
2. Stashed the overrides, reinstalled old deps on the **same machine**, regenerated **without the overrides** → diff vs committed baselines: `904 files changed, 1039 insertions(+), 1039 deletions(-)` — identical.

So with and without the four bumps the reference tree is byte-identical; the 904-vs-committed diff is pure environment drift (macOS / Node 20.20.2 vs the original Linux / Node 20.19.4) and pre-exists this PR. The four transitive dep bumps do not leak into SVG output.

`uuid` 14.0.1 (the one major bump, ESM-only) does not crash mermaid's module load or render path — the generator completed `exit 0` with `1394/1396 ok, 2 fail` (the 2 fails are the known frontier fixtures, same as on `main`).

The committed reference tree is left untouched; the code change is `package.json` + `package-lock.json` only; the Tier 2 rationale + audit are a README-only addition.

## Test plan

- [x] `npm install` reconciles `package-lock.json` under the overrides; 4 targets resolve to the pinned versions.
- [x] `node generate_ref.mjs --batch` succeeds; reference tree byte-identical to the without-overrides control.
- [ ] CI: `cargo run --bin sweep_all` and the `*_byte_exact.rs` suites unchanged (Rust tests read committed `tests/reference/**/*.svg` + `.mmd`, which this PR does not touch, so results are invariant by construction).

## Closes

Closes #140.

### DIFF
--- a/crates/mermaid-little/tests/support/README.md
+++ b/crates/mermaid-little/tests/support/README.md
@@ -49,6 +49,42 @@ tests/reference/
 └── ext_fixtures/<subsource>/<type>/<case>.svg
 ```
 
+## Upstream dep bumps (issue #140)
+
+A dependabot PR flagged five outdated deps here. Four are transitive deps of
+the generator with no bearing on its SVG output; they were bumped via
+`package.json#overrides` (commit `b35f808b`):
+
+- `dompurify` 3.4.1 → 3.4.12
+- `form-data` 4.0.5 → 4.0.6
+- `uuid` 11.1.0 → 14.0.1
+- `ws` 8.20.0 → 8.21.1
+
+The fifth, `mermaid` 11.14.0 → 11.15.0, is **deliberately not bumped**. A bump
+is not a version-string edit here — it regenerates and re-reviews the entire
+reference set, and 11.15.0 additionally breaks the generator on two fronts:
+
+- **Baseline drift**: 11.15.0 includes rendering changes (e.g. decimal
+  `autonumber` in sequence diagrams), so every baseline SVG would shift and
+  need re-approval, not just a re-run.
+- **jsdom generator breakage**: 11.15.0's stylesheet path hits the read-only
+  `CSSStyleSheet` getter in jsdom 25.0.1, which the generator's shims don't
+  cover. Fixing it means widening the shims, which in turn diverges the
+  generator's DOM environment from the Rust-side `stylis.rs` CSS parity
+  surface the reference set is meant to validate.
+
+So the call is: stay pinned on 11.14.0 (the version `tests/reference/VERSION`
+anchors the byte-exact comparison to) and revisit only when someone is
+prepared to re-baseline and re-shim together.
+
+The same five-package audit was applied to the sibling crates:
+
+- `crates/plantuml-little/tests/support/` — dep tree carries none of the five
+  (`npm ls` empty); only `@kookyleo/graphviz-anywhere-web`. No action.
+- `crates/d2-little/` — no support `package.json` with dependencies;
+  MathJax is vendored (`mathjax.js`), `elkjs` is resolved from the workspace
+  `engines` package. No action.
+
 ## Determinism
 
 Confirmed: same `.mmd` + same fixture path → identical SVG bytes across
--- a/crates/mermaid-little/tests/support/package-lock.json
+++ b/crates/mermaid-little/tests/support/package-lock.json
@@ -45,41 +45,10 @@
       "integrity": "sha512-jigsZK+sMF/cuiB7sERuo9V7N9jx+dhmHHnQyDSVdpZwVutaBu7WvNYqMDLSgFgfB30n452TP3vjDAvFC973mA==",
       "license": "MIT"
     },
-    "node_modules/@chevrotain/cst-dts-gen": {
-      "version": "12.0.0",
-      "resolved": "https://registry.npmjs.org/@chevrotain/cst-dts-gen/-/cst-dts-gen-12.0.0.tgz",
-      "integrity": "sha512-fSL4KXjTl7cDgf0B5Rip9Q05BOrYvkJV/RrBTE/bKDN096E4hN/ySpcBK5B24T76dlQ2i32Zc3PAE27jFnFrKg==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@chevrotain/gast": "12.0.0",
-        "@chevrotain/types": "12.0.0"
-      }
-    },
-    "node_modules/@chevrotain/gast": {
-      "version": "12.0.0",
-      "resolved": "https://registry.npmjs.org/@chevrotain/gast/-/gast-12.0.0.tgz",
-      "integrity": "sha512-1ne/m3XsIT8aEdrvT33so0GUC+wkctpUPK6zU9IlOyJLUbR0rg4G7ZiApiJbggpgPir9ERy3FRjT6T7lpgetnQ==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@chevrotain/types": "12.0.0"
-      }
-    },
-    "node_modules/@chevrotain/regexp-to-ast": {
-      "version": "12.0.0",
-      "resolved": "https://registry.npmjs.org/@chevrotain/regexp-to-ast/-/regexp-to-ast-12.0.0.tgz",
-      "integrity": "sha512-p+EW9MaJwgaHguhoqwOtx/FwuGr+DnNn857sXWOi/mClXIkPGl3rn7hGNWvo31HA3vyeQxjqe+H36yZJwYU8cA==",
-      "license": "Apache-2.0"
-    },
     "node_modules/@chevrotain/types": {
-      "version": "12.0.0",
-      "resolved": "https://registry.npmjs.org/@chevrotain/types/-/types-12.0.0.tgz",
-      "integrity": "sha512-S+04vjFQKeuYw0/eW3U52LkAHQsB1ASxsPGsLPUyQgrZ2iNNibQrsidruDzjEX2JYfespXMG0eZmXlhA6z7nWA==",
-      "license": "Apache-2.0"
-    },
-    "node_modules/@chevrotain/utils": {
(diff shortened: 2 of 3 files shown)


kookyleo (approved):
## Review: approving — the safety claim reproduced at zero byte differences, but the change is larger than the title says

**The central claim holds, and I verified it the hard way** rather than by reasoning about it. Two full trees on identical Node v20.19.4, differing only in dependency set — pre-PR lock vs this PR's lock — installed versions confirmed divergent (katex 0.16.45 vs 0.16.47, cytoscape 3.33.2 vs 3.34.0, parser 1.1.0 vs 1.2.0, langium present vs absent):

```
npm ci (pre-PR)  -> exit 0, 189 packages
npm ci (PR)      -> exit 0, 171 packages      <- lock IS consistent with package.json
generate_ref.mjs --batch --workers 6, both    -> exit 0, 1394 svg / 1396 total each
diff -rq preRef/ prRef/                       -> exit 0, 0 differing files
```

**Zero byte differences across all 1394 generated SVGs**, and that holds for the undeclared bumps too, not only the four declared ones. Drift against the committed reference tree is identical in both arms (894 files either way), so this PR neither causes nor worsens it.

Coverage is real, not incidental: 321 generated SVGs contain KaTeX output, 25 mindmap fixtures drive cytoscape, 53 gantt fixtures drive dayjs. `cargo test -p mermaid-little` exit 0, 1031 passed / 0 failed / 9 ignored.

**Security posture checked.** None of the four bumped packages carries an advisory at either the old or the new version — `form-data` 4.0.5 already post-dates CVE-2025-7783, `ws` 8.20.0 already post-dates CVE-2024-37890, `dompurify` 3.4.x is past the mXSS line. So this is hygiene, not a fix, which matches the Tier 1 framing. No yanked or deprecated releases. `uuid` 11.1.0 -> 14.0.1 crosses three majors and exceeds mermaid's declared `^11.1.0`, but the surface mermaid uses is intact at 14.0.1.

### Please fix in a follow-up — the description and README are now inaccurate

**1. The lockfile moves 17 packages, not 4.** The four overrides are real, but the lock was fully re-resolved: `@mermaid-js/parser` 1.1.0 -> 1.2.0, `cytoscape` 3.33.2 -> 3.34.0, `katex` 0.16.45 -> 0.16.47, `dayjs`, `@iconify/utils`, `ts-dedent`, `nwsapi`, `@chevrotain/types` 12.0.0 -> 11.1.2, plus 19 packages removed and 1 added. That accounts for the -235 lines: `@mermaid-js/parser@1.2.0` swapped `langium ^4.0.0` for `@chevrotain/types ~11.1.2`, dropping the whole langium/chevrotain/vscode-languageserver subtree, and `@iconify/utils@3.1.4` swapped `mlly` for `import-meta-resolve`.

None of that churn is required by the overrides. Control: pre-PR lock + this PR's `package.json`, `npm install --package-lock-only` changes exactly 5 packages (the 4 targets plus `hasown` 2.0.3 -> 2.0.4, pulled by `form-data@4.0.6`'s widened range) — a 36-line diff instead of 301. Either regenerate minimally or restate the scope.

**2. `README.md:26` and `:95` are now false.** Line 26 says the transitive `chevrotain@12` wants Node 22+; line 95 lists "chevrotain parser behaviour could theoretically differ" as a drift risk. This PR removes `chevrotain` from the tree entirely. The PR touches this exact file and adds 36 lines without correcting them — which is also the clearest sign the parser swap went unnoticed.

**3. `README.md:52-61` (added) says the four bumps have "no bearing on its SVG output" but does not mention katex, cytoscape or the parser moving.** True as written, incomplete as documentation.

### Other observations (no action needed here)

- The README's "Requires Node 20+" is wrong and unenforced: the generator crashes on Node 22 for both old and new deps (`TypeError: Cannot set property navigator` at `generate_ref.mjs:81`). It needs Node 20.x specifically, and no `engines` field says so. #153 addresses the crash itself.
- `package.json#overrides` is not mirrored into the lockfile root, so `npm ci` cannot detect drift between the two — they have to be reviewed together.
- The committed reference tree is not reproducible even on its own recorded environment: regenerating with the pre-PR dependency set on exactly the recorded Node v20.19.4 / Linux x86_64 still yields 894 files differing from the baselines. The byte-exact contract the reference tree exists to enforce is already broken upstream of this PR. Worth its own issue.
- Nothing in CI installs or exercises `crates/mermaid-little/tests/support` — only `plantuml-little`'s equivalent is wired in. Green CI here proves nothing about the generator; the differential run above is the only evidence that exists.
- `npm audit` reports one moderate finding, `mermaid@11.14.0` itself, deliberately retained per the Tier 2 decision. The package is `private: true` and never on a consumer path.

> Verification scope: the zero-diff result covers the 1394 fixtures that render today. A future fixture or a later mermaid bump could still reach a parser 1.1->1.2 or cytoscape 3.33->3.34 behavioural difference this corpus does not. I did not re-run the cross-crate `npm ls` audit asserted in the README, nor the cmark-gfm conformance suite (untouched by these files).

PR #151, opened 2026-07-30
fix(graphviz-rn): fix iOS New Architecture build (codegen name, .mm)

### LINKED ISSUE #136: [graphviz-anywhere/rn] iOS New Architecture build fails: codegen spec name mismatch + .m holding C++
> **Migrated from [kookyleo/graphviz-anywhere#8](https://github.com/kookyleo/graphviz-anywhere/issues/8)**, originally reported by **@boomsi** on 2026-07-25.
> GitHub cannot transfer issues across owners (user → org), so it was re-filed here by hand. The upstream repo has been subtree-merged into this one at `crates/graphviz-anywhere/`; the files referenced below are byte-identical between the two, so the report applies unchanged.
>
> **Triage result (see the "Verification" section at the bottom): items 1 and 2 are confirmed defects; items 3 and 4 did not reproduce and should not be "fixed" as written.**

---

## Original report

Building `@actrium/graphviz-anywhere-rn` (iOS) under the New Architecture (`RCT_NEW_ARCH_ENABLED=1`, RN 0.81/0.83) fails. It only surfaces when **iOS + New Arch + the GraphvizNative pod is installed** all hold — old architecture and Android don't trigger it (Android codegen name matches; Java has no C++ suffix issue).

### Issue 1 — codegen spec name has a stray "Native"

- `ios/GraphvizModule.h` declares `<NativeGraphvizNativeSpec>`; `ios/GraphvizModule.m` uses `NativeGraphvizNativeSpecJSI`.
- But with `codegenConfig.name = GraphvizNativeSpec`, RN's codegen produces protocol `NativeGraphvizSpec` / JSI `NativeGraphvizSpecJSI` (the rule strips the `Native` token from the name, then prefixes `Native`).
- `.h`/`.m` have one extra `Native` → `Cannot find protocol declaration for 'NativeGraphvizNativeSpec'; did you mean 'NativeGraphvizSpec'?`.
- Also: commit `08b49edf` "unify React Native package naming" changed `codegenConfig.name` from `GraphvizNativeSpec` to `GraphvizAnywhereNativeSpec` without updating `.h`/`.m`, which also breaks the import path (`.h` does `#import <GraphvizNativeSpec/...>` but codegen emits `GraphvizAnywhereNativeSpec/`).

### Issue 2 — implementation is `.m` but contains C++

- `ios/GraphvizModule.m` has `getTurboModule:` returning `std::make_shared<...>` (C++) and `#import`s the codegen spec header (C++).
- Compiled as ObjC (`.m`) → no C++ STL include path → `'functional'/'cassert'/'iosfwd'/'optional'/'utility' file not found`, plus C++ syntax errors.
- Fix: rename `GraphvizModule.m` → `GraphvizModule.mm`. `podspec` `source_files = "*.{h,m,mm}"` already covers `.mm`.

### Issue 3 — xcframework name mismatch

- `podspec` references `Frameworks/Graphviz.xcframework`, but the on-disk file is `GraphvizApi.xcframework`.
- `build-ios.sh` outputs `Graphviz.xcframework` (line ~173) and `prepare-native.js` stages it as `Graphviz.xcframework` — so the present `GraphvizApi.xcframework` looks like a non-standard/incomplete artifact.

### Issue 4 — xcframework ships no headers

- `GraphvizApi.xcframework/<slice>/` contains only `libgraphviz_api.a`, **no `graphviz_api.h`**.
- `build-ios.sh` does copy `graphviz_api.h` into slices (lines ~133/150) and `output/ios/`, but `output/ios/` doesn't exist (build-ios.sh never ran successfully), so the staged xcframework has no headers.
- The header source is `capi/graphviz_api.h` (cbindgen, self-contained — only `stddef.h`/`stdint.h`).
- Fix: run `build-ios.sh` to produce a complete `Graphviz.xcframework` (with Headers), or include `capi/graphviz_api.h` in the xcframework.

### Reproduce

1. New Arch host app (`RCT_NEW_ARCH_ENABLED=1`).
2. `pod 'GraphvizNative', :path => '...'`.
3. `pod install` + build iOS → fails at `GraphvizModule.h`/`.mm`, then missing `graphviz_api.h`.

### Suggested minimal fixes (verified locally by the reporter — BUILD SUCCEEDED)

- `.h`/`.mm`: `NativeGraphvizNativeSpec` → `NativeGraphvizSpec`, and `…SpecJSI` likewise.
- `git mv GraphvizModule.m GraphvizModule.mm`.
- `podspec`: `Graphviz.xcframework` → `GraphvizApi.xcframework` (or fix the build to produce `Graphviz.xcframework`).
- Ship `graphviz_api.h` inside the xcframework (run build-ios.sh, or include `capi/graphviz_api.h`).
- Reconcile `codegenConfig.name` with `.h`/`.mm` (currently `GraphvizNativeSpec` + `NativeGraphvizSpec` works).

### Notes

- Only `codegenConfig.name` and the iOS sources are inconsistent; the runtime module name (`RCT_EXPORT_MODULE(GraphvizNative)` / `TurboModuleRegistry.getEnforcing('GraphvizNative')`) is consistent and unrelated.

---

## Verification

Checked by reading the sources in `crates/graphviz-anywhere/packages/react-native/`. **No iOS build was run** (no macOS/Xcode available) — the reporter states they verified their patch builds successfully. Upstream `kookyleo/graphviz-anywhere` was cloned and diffed against the in-tree subtree copy: the files below are identical, so this applies to both.

### Item 1 — CONFIRMED (two independent mismatches)

- `ios/GraphvizModule.h:19` declares `NativeGraphvizNativeSpec`, `ios/GraphvizModule.m:156` instantiates `NativeGraphvizNativeSpecJSI`. The JS spec file is `src/NativeGraphviz.ts`, from which RN codegen derives `NativeGraphvizSpec` / `NativeGraphvizSpecJSI` — the ObjC side carries one extra `Native`.
- `ios/GraphvizModule.h:14` imports `<GraphvizNativeSpec/GraphvizNativeSpec.h>`, but `package.json:87` sets `codegenConfig.name` to `GraphvizAnywhereNativeSpec`, so codegen emits a `GraphvizAnywhereNativeSpec/` directory instead.
- The cited commit checks out: `08b49edf` ("refactor: unify React Native package naming", 2026-03-15) changed exactly `"name": "GraphvizNativeSpec"` → `"GraphvizAnywhereNativeSpec"` in `package.json` and touched no ObjC source.

### Item 2 — CONFIRMED

`ios/GraphvizModule.m:154-156` declares `- (std::shared_ptr<facebook::react::TurboModule>)getTurboModule:` returning `std::make_shared<...>` — C++ in a `.m` translation unit. `ios/GraphvizNative.podspec:20` already has `source_files = "*.{h,m,mm}"`, so renaming the file is sufficient; the podspec needs no change.

### Item 3 — NOT REPRODUCED

The string `GraphvizApi.xcframework` does not appear anywhere in this repository. The pipeline is internally consistent on `Graphviz.xcframework`:
- `crates/graphviz-anywhere/scripts/build-ios.sh:167,173,176` creates and verifies `Graphviz.xcframework`
- `packages/react-native/scripts/prepare-native.js:43,112` copies it to `ios/Frameworks/Graphviz.xcframework`
- `ios/GraphvizNative.podspec:28` declares `vendored_frameworks = "Frameworks/Graphviz.xcframework"`

The release pipeline agrees too: `.github/workflows/graphviz-release.yml` packages the whole `output/ios/` tree into `graphviz-native-ios.tar.gz`, which contains `Graphviz.xcframework`. The reporter's `GraphvizApi.xcframework` is not produced by any code path here. **Changing the podspec to match it would break the supported build.**

### Item 4 — NOT REPRODUCED as a repository defect

`scripts/build-ios.sh:170-173` passes `-headers "${HEADER_DIR}"` for both the device and simulator slices, and `:150` populates `HEADER_DIR` with `graphviz_api.h` just beforehand. `xcodebuild -create-xcframework` places those under `<slice>/Headers/`, which is exactly where `ios/GraphvizNative.podspec:31-32` points `HEADER_SEARCH_PATHS`. The reporter's missing headers follow from item 3 — their xcframework was not produced by `build-ios.sh`, which they note themselves ("build-ios.sh never ran successfully").

### Suggested scope

Fix items 1 and 2 only:

1. `ios/GraphvizModule.h` / `.m`: `NativeGraphvizNativeSpec` → `NativeGraphvizSpec`, `NativeGraphvizNativeSpecJSI` → `NativeGraphvizSpecJSI`.
2. Reconcile the import path with `codegenConfig.name` (either revert `package.json:87` to `GraphvizNativeSpec`, or update the `#import` to `GraphvizAnywhereNativeSpec/`).
3. `git mv ios/GraphvizModule.m ios/GraphvizModule.mm`.

Items 3 and 4 need no code change, but the fact that a consumer got this far with a hand-made xcframework suggests `prepare-native.js` should fail loudly when `output/ios/Graphviz.xcframework` is absent rather than leaving a partly-staged `ios/Frameworks/`.

Since none of this can be verified without macOS + Xcode, the fix needs an iOS New Arch build check before merging.

boomsi:
Fixes #136 (items 1 & 2 per the issue's triage).

## Problem

iOS New Architecture build of `@actrium/graphviz-anywhere-rn` (`RCT_NEW_ARCH_ENABLED=1`) fails. Only surfaces with iOS + New Arch + the GraphvizNative pod installed (Android codegen name already matches; old arch unaffected).

## Fix — items 1 & 2 only (per #136 triage scope)

**Item 1 — codegen spec name has a stray "Native"**
- `package.json`: `codegenConfig.name` `GraphvizAnywhereNativeSpec` → `GraphvizNativeSpec` — reconciles `.h` `#import <GraphvizNativeSpec/…>` with the codegen output directory (commit `08b49edf` renamed `codegenConfig.name` without updating the ObjC sources).
- `GraphvizModule.h` / `.mm`: `NativeGraphvizNativeSpec` → `NativeGraphvizSpec`; `NativeGraphvizNativeSpecJSI` → `NativeGraphvizSpecJSI`. RN codegen strips the `Native` token from `codegenConfig.name` then prefixes `Native`, so the spec is `NativeGraphvizSpec`/`…SpecJSI`; the ObjC side carried one extra `Native` → `Cannot find protocol declaration for 'NativeGraphvizNativeSpec'`.

**Item 2 — implementation is `.m` but contains C++**
- `git mv GraphvizModule.m → GraphvizModule.mm`. `getTurboModule:` returns `std::make_shared<…>` and `#import`s the codegen spec header (C++); compiled as `.m` it had no C++ STL include path. `podspec` `source_files = "*.{h,m,mm}"` already covers `.mm`, so **the podspec needs no change**.

## Why items 3 & 4 are not fixed

Per #136 triage, neither reproduces as a repository defect:
- The standard pipeline (`scripts/build-ios.sh` → `Graphviz.xcframework` → `scripts/prepare-native.js`) is internally consistent on `Graphviz.xcframework`. The reporter's `GraphvizApi.xcframework` was a hand-made artifact (`build-ios.sh` never ran). Changing the podspec to match it would **break the supported build**.
- Missing headers follow from the hand-made artifact; `build-ios.sh` does stage `graphviz_api.h` under each slice's `Headers/`.

## Test plan

- [x] Built a standard `Graphviz.xcframework` via `scripts/build-ios.sh` (3 slices: iphoneos-arm64, iphonesimulator-arm64, iphonesimulator-x86_64; per-slice `Headers/graphviz_api.h` + `libgraphviz_api.a`).
- [x] Staged via `node scripts/prepare-native.js` → `ios/Frameworks/Graphviz.xcframework`.
- [x] `pod install` + iOS New Arch build (`RCT_NEW_ARCH_ENABLED=1`) **SUCCEEDED** in the host app against the standard xcframework.
- [x] Runtime: dot graph renders correctly.

Only 3 source files are committed (`build/` and `output/` are gitignored).

Note: the original `kookyleo/graphviz-anywhere` is archived; `crates/graphviz-anywhere/` is the in-repo copy and supramark is the live maintainer, so this PR targets `Actrium/supramark#136` directly.

### DIFF
--- a/crates/graphviz-anywhere/packages/react-native/ios/GraphvizModule.h
+++ b/crates/graphviz-anywhere/packages/react-native/ios/GraphvizModule.h
@@ -16,7 +16,7 @@
 
 @interface GraphvizModule : NSObject <RCTBridgeModule
 #ifdef RCT_NEW_ARCH_ENABLED
-  , NativeGraphvizNativeSpec
+  , NativeGraphvizSpec
 #endif
 >
 
--- a/crates/graphviz-anywhere/packages/react-native/ios/GraphvizModule.mm
+++ b/crates/graphviz-anywhere/packages/react-native/ios/GraphvizModule.mm
@@ -153,7 +153,7 @@ RCT_EXPORT_METHOD(getVersion:(RCTPromiseResolveBlock)resolve
 #ifdef RCT_NEW_ARCH_ENABLED
 - (std::shared_ptr<facebook::react::TurboModule>)getTurboModule:
     (const facebook::react::ObjCTurboModule::InitParams &)params {
-    return std::make_shared<facebook::react::NativeGraphvizNativeSpecJSI>(params);
+    return std::make_shared<facebook::react::NativeGraphvizSpecJSI>(params);
 }
 #endif
 
--- a/crates/graphviz-anywhere/packages/react-native/macos/GraphvizModule.h
+++ b/crates/graphviz-anywhere/packages/react-native/macos/GraphvizModule.h
@@ -16,7 +16,7 @@
 
 @interface GraphvizModule : NSObject <RCTBridgeModule
 #ifdef RCT_NEW_ARCH_ENABLED
-  , NativeGraphvizNativeSpec
+  , NativeGraphvizSpec
 #endif
 >
 
--- a/crates/graphviz-anywhere/packages/react-native/macos/GraphvizModule.mm
+++ b/crates/graphviz-anywhere/packages/react-native/macos/GraphvizModule.mm
@@ -1,5 +1,5 @@
 /*
- * GraphvizModule.m (macOS)
+ * GraphvizModule.mm (macOS)
  *
  * React Native native module implementation for macOS.
  * Manages a singleton Graphviz context and dispatches rendering
@@ -156,7 +156,7 @@ RCT_EXPORT_METHOD(getVersion:(RCTPromiseResolveBlock)resolve
 #ifdef RCT_NEW_ARCH_ENABLED
 - (std::shared_ptr<facebook::react::TurboModule>)getTurboModule:
     (const facebook::react::ObjCTurboModule::InitParams &)params {
-    return std::make_shared<facebook::react::NativeGraphvizNativeSpecJSI>(params);
+    return std::make_shared<facebook::react::NativeGraphvizSpecJSI>(params);
 }
 #endif
 
--- a/crates/graphviz-anywhere/packages/react-native/package.json
+++ b/crates/graphviz-anywhere/packages/react-native/package.json
@@ -84,7 +84,7 @@
     ]
   },
   "codegenConfig": {
-    "name": "GraphvizAnywhereNativeSpec",
+    "name": "GraphvizNativeSpec",
     "type": "modules",
     "jsSrcsDir": "src",
     "android": {
--- a/crates/graphviz-anywhere/packages/react-native/scripts/test-ios.sh
+++ b/crates/graphviz-anywhere/packages/react-native/scripts/test-ios.sh
@@ -9,7 +9,7 @@
 #
 #   1. uses the prebuilt Graphviz static lib produced by
 #      scripts/build-macos.sh (output/macos-universal/lib/libgraphviz_api.a),
-#   2. compiles the ObjC bridge (.m) + the ObjC++ XCTest (.mm) into a
+#   2. compiles the ObjC++ bridge (.mm) + the ObjC++ XCTest (.mm) into a
 #      macOS .xctest bundle against the minimal RN header stubs in
 #      ios/__tests__/stubs,
 #   3. runs it with `xcrun xctest`.
@@ -54,12 +54,12 @@ BUNDLE="$OUT/GraphvizModuleTests.xctest"
 mkdir -p "$BUNDLE/Contents/MacOS"
 BIN="$BUNDLE/Contents/MacOS/GraphvizModuleTests"
 
-# The bridge is Objective-C (.m); the test is Objective-C++ (.mm). Compile
-# each TU to an object file with the right language, then link them.
-clang -g -O0 -fobjc-arc -x objective-c \
+# The bridge and the test are both Objective-C++ (.mm). Compile each TU
+# with clang++ -x objective-c++, then link them.
+clang++ -g -O0 -fobjc-arc -std=c++17 -x objective-c++ \
   -isysroot "$SDK" \
   -I"$IOS_DIR" -I"$NATIVE_INCLUDE" -I"$STUBS" \
-  -c "$IOS_DIR/GraphvizModule.m" \
+  -c "$IOS_DIR/GraphvizModule.mm" \
   -o "$OUT/GraphvizModule.o"
 
 clang++ -g -O0 -fobjc-arc -std=c++17 -x objective-c++ \


kookyleo (approved):
## Review: approving — the three New Arch names line up, and the `.m` -> `.mm` rename is complete

Verified against source rather than the description, and cross-checked with the four sibling RN packages in this repo that already build with the same scheme.

**Name chain after this PR** — TS spec `src/NativeGraphviz.ts` -> codegen protocol `NativeGraphvizSpec` (`ios/GraphvizModule.h:19`), JSI class `NativeGraphvizSpecJSI` (`ios/GraphvizModule.mm:156`), umbrella header `#import <GraphvizNativeSpec/GraphvizNativeSpec.h>` (`ios/GraphvizModule.h:14`) matching `codegenConfig.name` (`package.json:87`). `mermaid-little` and `supramark-markdown` have the identical shape (`NativeSupramarkMermaid.ts` -> `NativeSupramarkMermaidSpec` + umbrella `SupramarkMermaidNativeSpec`), so graphviz-anywhere is no longer the odd one out.

**Runtime lookup is untouched and consistent**: `GraphvizNative` is identical across `RCT_EXPORT_MODULE`, `TurboModuleRegistry.getEnforcing`, `NativeModules.*`, the Java `getName()`, and the podspec. No "compiles but module not found" risk.

**Rename is complete**: no `.m` survives anywhere under the package; the podspec glob `*.{h,m,mm}` already covered `.mm`, so no podspec change was needed. The C header `capi/graphviz_api.h:16` has the `extern "C"` guards, so compiling the bridge as ObjC++ will not break linkage against the C static library.

**`scripts/test-ios.sh` is repaired, not weakened**: `set -euo pipefail` intact, final `exec xcrun xctest "$BUNDLE"` intact, no swallowed exit codes. Commit 1 left the script pointing at the deleted `.m`; commit 2 fixes it.

### Nits (non-blocking)

- `ios/GraphvizModule.mm:2` — the header comment still reads `GraphvizModule.m`. The macOS twin was updated in commit 2, so the two now disagree, and this is the only surviving `.m` string in the repo.
- The rule stated in the commit messages ("codegen strips the `Native` token from `codegenConfig.name` then prefixes `Native`") is not how RN codegen works — the protocol name derives from the JS spec filename, while `codegenConfig.name` only names the generated library and its umbrella header. The two coincide here. Worth correcting so a future rename of `codegenConfig.name` does not "fix" the protocol back into a broken state.
- `codegenConfig.name` moves back from `GraphvizAnywhereNativeSpec` to `GraphvizNativeSpec`, which steps away from the `<PackageName>NativeSpec` convention the four siblings follow. Resolving the mismatch by updating the two `#import` lines instead would have kept the convention. Both compile; flagging so the next naming sweep updates the imports too.
- The macOS half cannot be built by anything in the repo today: the only podspec is `ios/GraphvizNative.podspec` with `source_files = "*.{h,m,mm}"` relative to `ios/`, and no macOS slice is produced. The change is correct but currently unreachable.

### CI blind spot (separate from this PR)

Nothing in CI runs RN codegen, compiles `GraphvizModule.mm`, or runs `pod install` for this pod. `native-host-tests` explicitly excludes graphviz-anywhere; the `macos-14` jobs build the Rust example and the Graphviz static library, never the ObjC bridge; `test-ios.sh` is not referenced by any workflow. That is why #136 could land green, and why commit 1 could break the only iOS test script without CI noticing. A `macos-14` job running `scripts/test-ios.sh` would be the minimum gate, ideally a second one with `-DRCT_NEW_ARCH_ENABLED` against real codegen output.

> Verification scope, to be explicit: this is a Linux machine with no Xcode, so **no iOS or macOS build was performed**. The name alignment is established by reading sources and comparing against working sibling packages, not by running codegen. The device build and runtime render reported in the PR description are relayed, not confirmed. `test-ios.sh` should be run once on a Mac to confirm commit 2's repair, since nothing else will.

PR #153, opened 2026-07-30
fix(mermaid-little): guard globalThis assignment for read-only getters (Node >=21)

### LINKED ISSUE #152: mermaid-little tests/support generator crashes on Node ≥21 (globalThis.navigator is a read-only getter)
## Problem

`crates/mermaid-little/tests/support/generate_ref.mjs` crashes at pre-init on Node ≥ 21, so the reference SVG generator cannot run at all on modern Node. `tests/reference/VERSION` records the supported generator as Node `v20.19.4`; anything newer hits the failure below.

## Repro

```
$ node --version
v23.11.0

$ cd crates/mermaid-little/tests/support
$ node generate_ref.mjs --batch
[crash pre-init] TypeError: Cannot set property navigator of #<Object> which has only a getter
    at file:///.../generate_ref.mjs:81:41
    at ModuleJob.run (node:internal/modules/esm/module_job:274:25)
    ...
```

Confirmed also reproduces on the current `main` (independent of #149 / the dep-bump work).

## Root cause

`generate_ref.mjs:81`:

```js
for (const k of [ ... 'navigator', ... ]) {
  if (W[k] !== undefined) globalThis[k] = W[k];   // line 81
}
```

[Node 21 added a built-in `globalThis.navigator`](https://nodejs.org/api/globals.html#navigator) implemented as a read-only, non-configurable getter. ESM is strict mode, so assigning to a getter-only property throws `TypeError`. On Node 20 `globalThis.navigator` did not exist, so the assignment succeeded and copied jsdom's `navigator` over.

## Impact

- Cannot regenerate the `tests/reference/` baselines locally on Node ≥ 21.
- Anyone on a modern Node (the default `node` Homebrew formula is now 23.x) has to install `node@20` just to run the generator.
- Byte-exact verification work (e.g. #149's control run) is blocked on the default toolchain.

## Suggested fix direction (not implemented)

Guard the assignment so read-only globals don't crash it. Smallest change:

```js
if (W[k] !== undefined) {
  try { globalThis[k] = W[k]; } catch {}
}
```

or, more deliberately, skip keys whose existing `globalThis` descriptor is a non-writable getter:

```js
if (W[k] !== undefined) {
  const desc = Object.getOwnPropertyDescriptor(globalThis, k);
  if (!desc || desc.writable || desc.set) globalThis[k] = W[k];
}
```

Either keeps the Node 20 behavior intact (jsdom's `navigator` still gets installed there) while not throwing on Node ≥ 21. The `tests/reference/VERSION` `generator_node` field should also be bumped once the generator is confirmed green on a modern Node.

## Out of scope

The 904-vs-committed SVG diff observed when regenerating on macOS / Node 20.20.2 is environment drift (font/jsdom metrics), not caused by this crash and not addressed here.

mybee:
## Summary

Fixes #152.

`crates/mermaid-little/tests/support/generate_ref.mjs` crashed at pre-init on Node >= 21:

```
[crash pre-init] TypeError: Cannot set property navigator of #<Object> which has only a getter
    at .../generate_ref.mjs:81
```

The loop that copies jsdom window globals onto `globalThis` assigns `globalThis.navigator = W.navigator`. Node 21+ ships a built-in `globalThis.navigator` as a read-only, non-configurable getter, and ESM is strict mode, so the assignment throws. Only Node 20 (`tests/reference/VERSION` records `v20.19.4`) could run the generator.

## Fix

Wrap the assignment in `try/catch`:

```js
if (W[k] !== undefined) {
  try { globalThis[k] = W[k]; } catch {}
}
```

On Node 20 the assignment still succeeds (jsdom's `navigator` wins). On Node >= 21 the assignment is skipped, leaving Node's native `navigator` in place.

## Verification — zero output drift

The concern with skipping `navigator` is that mermaid might read a bare `globalThis.navigator` at module scope and branch on it differently than under jsdom's navigator. Proven not to be the case by direct tree comparison:

1. Regenerated the full 1394-fixture reference tree on **Node 23.11.0** with the fix → archived.
2. Regenerated on **Node 20.20.2** (no fix needed) → archived.
3. `diff -rq` between the two trees: **0 differing files**.

So the Node 23 output is byte-identical to the Node 20 output. Mermaid reads `navigator` via the jsdom `window` (`W`), not `globalThis`, so leaving Node's native global in place does not affect SVG rendering. Generator completed `exit 0`, `1394/1396 ok, 2 fail` (the 2 fails are the known frontier fixtures, same as on `main`).

The committed `tests/reference/` baselines are untouched; this PR contains only the one-line guard in `generate_ref.mjs`.

## Test plan

- [x] `node generate_ref.mjs --batch` on Node 23.11.0: no pre-init crash, `1394/1396 ok`.
- [x] `node generate_ref.mjs --batch` on Node 20.20.2: still works, `1394/1396 ok`.
- [x] `diff -rq` Node 23 tree vs Node 20 tree: 0 differing files.
- [ ] CI: `cargo run --bin sweep_all` unchanged (Rust reads committed `tests/reference/**/*.svg` + `.mmd`, which this PR does not touch, so results are invariant by construction).

## Note

The `generator_node` field in `tests/reference/VERSION` still records `v20.19.4`. Bumping that field is intentionally left for a follow-up — it should be updated together with a CI toolchain bump so the recorded version matches what CI actually runs.


kookyleo (dismissed):
## Review: approving — the "silently skipped polyfill" smell is real, but proven to be a byte-exact no-op

The obvious way this fix goes wrong is that the guard swallows the assignment, the polyfill never lands, and the generator silently emits *different* reference SVGs instead of crashing. That is exactly what happens — and it turns out not to matter, which I established by measurement rather than by argument.

**The skip is real.** On Node v22.21.1, minimal repro faithful to `generate_ref.mjs:62-81`:

```
OLD code -> exit 1, TypeError: Cannot set property navigator of #<Object> which has only a getter
NEW code -> exit 0, globalThis.navigator === W.navigator  ->  false     <- polyfill IS skipped
```

**And it is a no-op.** Four-arm reference regeneration over all 1396 fixtures, holding Node version, OS, fonts and dependency set constant and varying *only* whether jsdom's navigator is force-installed:

```
arm C  node 22.21.1, post-fix as merged        -> 1394 svg / 2 fail
arm D  node 22.21.1, post-fix + navigator forced -> 1394 svg / 2 fail
diff -rq C D  -> exit 0, 0 differing files
```

That is a tighter control than the PR's own (which compared Node 23 against Node 20, conflating node version with navigator). A global-navigator read spy over a real render confirms the mechanism: **0 reads** of `globalThis.navigator` — mermaid reaches navigator through the jsdom `window`, as the description claims.

A fifth arm additionally clears the dependency bump (old lock vs new lock on the same code): 0 differing files. Drift against the committed baselines is constant at 901 files across every arm, so this PR adds none.

`cargo test -p mermaid-little`: exit 0, 1031 passed / 0 failed / 9 ignored. The 1394 committed reference SVGs are untouched, which is correct rather than a missing regeneration.

**Merge order relative to #149:** not a conflict — a literally shared commit. `b35f808b` is identical in both PRs (same SHA, same blobs), so this PR is stacked on that dependency bump. #149 has been merged first, so what remains here is the single 9-line generator fix.

### Please fix in a follow-up

**1. `generate_ref.mjs:81-82` — the comment is factually wrong and steers the next maintainer away from the correct fix.** It says Node ships `navigator` as a read-only, *non-configurable* getter. Measured on Node v22.21.1: `get=true set=false configurable=true`. Because it *is* configurable, `Object.defineProperty` works and installs jsdom's navigator faithfully — verified. A shape that preserves Node 20 semantics exactly on every version:

```js
try { globalThis[k] = W[k]; }
catch { Object.defineProperty(globalThis, k, { value: W[k], writable: true, configurable: true }); }
```

The same claim is in the commit message and the PR body.

**2. `generate_ref.mjs:88` — `catch {}` silently swallows across all 17 keys with no diagnostic.** Today only `navigator` is affected (the other 16 are absent on Node 22, so plain assignment works). If a future Node ships a read-only `location`, `screen` or `CSS`, the shim degrades silently and the generator emits wrong SVGs with no signal. Warn on the caught key, or scope the catch to `navigator`.

**3. The fix is half-applied.** 18 sibling scripts in the same directory still carry the identical unguarded loop and still crash on Node >= 21: `probe_render.mjs:14`, `dagre_debug.mjs:29`, and 16 `test_13[45]_*.mjs`. They are committed debug tooling, not CI-run, so this is cosmetic — but the stated goal of running the generator tooling on modern Node is only half met.

**4. `tests/reference/VERSION` still records `generator_node: v20.19.4`**, now understating supported Node. Self-disclosed as follow-up, which is fine.

### Observation, not this PR's problem

The committed reference tree is not reproducible even on its own recorded environment: Node 20.19.4 with the pre-PR lockfile and unmodified `main` code still differs from the baselines in ~901 files. Same drift #149 and #152 both report. The byte-exact contract the reference tree exists to enforce is already broken upstream. Worth its own issue.

Also worth noting: no workflow installs `crates/mermaid-little/tests/support` or runs `generate_ref.mjs` — the `mermaid-little` job only runs `cargo test` against the committed SVGs. So both the lockfile change and this fix have zero CI blast radius, and green CI here proves nothing about either. The runs above are the only evidence.

> Verification scope: tested on Node v22.21.1 and v20.19.4 on Linux. Node 23.x — the version in the report — was not exercised, though Node 22 shows the identical read-only-getter behaviour so the conclusion carries. macOS not tested.

kookyleo (dismissed):
Re-approving: the previous approval was auto-dismissed when #149 landed and changed the merge base. Findings and verification unchanged — see the full review above.

kookyleo (approved):
Re-approving on the updated head (branch refreshed against current main; the earlier approval was auto-dismissed when #149 changed the merge base). Findings and verification unchanged — see the full review above.

PR #155, opened 2026-07-30
fix(footnote,emoji): re-export feature config helpers from index.ts

### LINKED ISSUE #146: footnote / emoji feature packages don't re-export their config helpers from index.ts
## Summary

`packages/features/main/footnote/src/index.ts` and `packages/features/main/emoji/src/index.ts` do not re-export the feature-config helper functions/types that `feature.ts` defines for them, unlike every other sibling feature package that follows the same `makeFeatureConfigHelpers` pattern (math, gfm, core-markdown, definition-list, html-page, map).

## Details

Both `footnote/src/feature.ts` and `emoji/src/feature.ts` define and export the standard config-helper set via `makeFeatureConfigHelpers`:

```ts
// footnote/src/feature.ts
export type FootnoteFeatureConfig = FeatureConfigWithOptions<FootnoteFeatureOptions>;
const footnoteHelpers = makeFeatureConfigHelpers<FootnoteFeatureOptions>(...);
export const createFootnoteFeatureConfig = footnoteHelpers.create;
export const getFootnoteFeatureOptions = footnoteHelpers.getOptions;
```

```ts
// emoji/src/feature.ts
const emojiHelpers = makeFeatureConfigHelpers<EmojiFeatureOptions>('@supramark/feature-emoji');
export const createEmojiFeatureConfig = emojiHelpers.create;
export const getEmojiFeatureOptions = emojiHelpers.getOptions;
```

But their `index.ts` (the package's public entry point) only re-exports the feature object and examples:

```ts
// footnote/src/index.ts
export { footnoteFeature } from './feature.js';
export { footnoteExamples } from './examples.js';
```

```ts
// emoji/src/index.ts
export { emojiFeature } from './feature.js';
export { emojiExamples } from './examples.js';
```

Compare with `math/src/index.ts`, which does it correctly:

```ts
export {
  mathFeature,
  type MathFeatureOptions,
  type MathFeatureConfig,
  createMathFeatureConfig,
  getMathFeatureOptions,
} from './feature.js';
```

## Impact

A consumer importing from the package root (`@supramark/feature-footnote` or `@supramark/feature-emoji`, which is what `main`/`types`/`exports` in each `package.json` point at per the project's TS path convention) cannot reach `createFootnoteFeatureConfig`, `getFootnoteFeatureOptions`, `createEmojiFeatureConfig`, or `getEmojiFeatureOptions` — even though they exist and are documented in each feature's own `documentation.api.functions` metadata. This is an inconsistent/broken public API surface relative to every other feature package.

## Suggested fix

Add the missing re-exports to both `index.ts` files, mirroring the pattern already used in `math/src/index.ts`, `gfm/src/index.ts`, `core-markdown/src/index.ts`, `definition-list/src/index.ts`, `html-page/src/index.ts`, and `map/src/index.ts`.

## How this was found

Noticed incidentally while doing an unrelated pass over these packages, not investigated further — flagging for someone with more context on the feature-package conventions to fix and verify against `scripts/features-lint.ts` / `validateFeature()`.

mybee:
## Summary

- `footnote` and `emoji` feature packages define their config helpers (`createFootnoteFeatureConfig` / `getFootnoteFeatureOptions` and emoji equivalents) plus `*FeatureOptions` / `*FeatureConfig` types in `feature.ts`, but their package entry `index.ts` did not re-export them.
- Consumers importing from the package root (`@supramark/feature-footnote` / `@supramark/feature-emoji`) could not reach these helpers — inconsistent with every sibling feature (`math`, `gfm`, `core-markdown`, `definition-list`, `html-page`, `map`).
- Mirrors the re-export pattern already used in `packages/features/main/math/src/index.ts`.

## Test plan

- [x] `bun run build` (tsc --emitDeclarationOnly) passes for both packages
- [x] `bun run features:lint` — 21/21 features pass

Closes #146

### DIFF
--- a/packages/features/main/emoji/src/index.ts
+++ b/packages/features/main/emoji/src/index.ts
@@ -4,7 +4,13 @@
  * @packageDocumentation
  */
 
-export { emojiFeature } from './feature.js';
+export {
+  emojiFeature,
+  type EmojiFeatureOptions,
+  type EmojiFeatureConfig,
+  createEmojiFeatureConfig,
+  getEmojiFeatureOptions,
+} from './feature.js';
 export { emojiExamples } from './examples.js';
 
 // Note: the Emoji Feature uses the standard SupramarkTextNode.
--- a/packages/features/main/footnote/src/index.ts
+++ b/packages/features/main/footnote/src/index.ts
@@ -4,7 +4,13 @@
  * @packageDocumentation
  */
 
-export { footnoteFeature } from './feature.js';
+export {
+  footnoteFeature,
+  type FootnoteFeatureOptions,
+  type FootnoteFeatureConfig,
+  createFootnoteFeatureConfig,
+  getFootnoteFeatureOptions,
+} from './feature.js';
 export { footnoteExamples } from './examples.js';
 
 // Re-export core types (for user convenience)


kookyleo (approved):
## Review: approving — the bug was real at both type and runtime level, and the fix is exactly the house pattern

Verified independently rather than from the description.

**The gap was real.** Both packages set `main`, `types` and `exports["."]` to `./src/index.ts` with no secondary subpath, so `index.ts` is the only way in — and the helpers (`create*FeatureConfig`, `get*FeatureOptions`, `*FeatureOptions`, `*FeatureConfig`) were defined in `feature.ts` and never re-exported. The packages' own `documentation.api` metadata already published import examples that could not work: `footnote/src/feature.ts:399` reads `import { createFootnoteFeatureConfig } from '@supramark/feature-footnote';`.

**Negative control:** reverting the two files and re-running the probe gives 8 type errors (`TS2305`/`TS2724`) plus a runtime `SyntaxError: Export named 'getEmojiFeatureOptions' not found`. Re-applied, both clean.

**Correctness of the re-exports:** all four symbols per package, nothing left behind. The value/type split is right — interfaces and type aliases get inline `type`, the two `const` bindings are exported as values. Type-checked under `isolatedModules: true` + `verbatimModuleSyntax: true`, both stricter than `tsconfig.base.json`. No name collisions, no `export *` that could merge-shadow them, no import cycle.

**House convention:** swept all 21 feature packages — after this PR the codebase is 100% consistent on the `export { xFeature, type XFeatureOptions, type XFeatureConfig, createX…, getX… } from './feature.js'` shape. No third package carries the same bug.

**Probe:** imported only through the public entries and executed — value-callability, assignability to the exported config aliases, the `enabled` default path, the disabled path, a detached call (proving the arrow-property typing in `packages/core/src/feature.ts:14` keeps `this` a non-issue), `getOptions` round-trip, and the `undefined`-config path. `tsc` exit 0, runtime exit 0.

Tests: `feature-footnote` 5 pass / 0 fail, `feature-emoji` 4 pass / 0 fail, both `build` (`tsc --emitDeclarationOnly`) exit 0, `features:lint` 21/21 passed, `lint:types` total 0, `check-cjk` clean.

### Follow-ups (non-blocking, not for this PR)

- No test in the repo imports a feature package through its public entry — the existing suites import `../src/feature` directly, so neither they nor `features-lint` would have caught this, or catch a regression. A features-lint rule diffing `feature.ts`'s exported `create*FeatureConfig`/`get*FeatureOptions` against `index.ts` would close the class.
- `admonition` and `weather` never call `makeFeatureConfigHelpers` at all, so they have no typed config helper API to export. Different gap from #146, worth its own issue.
- `tsconfig.base.json` is missing `paths` entries for `@supramark/feature-plantuml`, `@supramark/feature-d2` and `@supramark/feature-card-vison`; they resolve via workspace symlinks, so cosmetic.

> Verification scope: full `bun run build` could not complete here — three wasm targets fail on a local global linker setting unrelated to this change. All 36 TypeScript packages built clean, and CI is green on this head.

PR #156, opened 2026-07-30
refactor(weather): dedup getMockWeather across web and RN runtimes

### LINKED ISSUE #147: weather feature: getMockWeather duplicated between web and RN runtimes, condition representation differs
## Summary

`packages/features/containers/weather/src/runtime.web.tsx` and `packages/features/containers/weather/src/runtime.rn.tsx` each define their own copy of `getMockWeather(location, units)` with the same hashing/derivation logic duplicated verbatim, except for one field that differs silently between the two copies: the `condition` value.

## Details

Both files compute temperature, humidity and wind identically by hashing the `location` string's character codes:

```ts
// runtime.web.tsx (around line 57)
function getMockWeather(location: string, units: 'metric' | 'imperial' = 'metric') {
  const hash = location.split('').reduce((acc, c) => acc + c.charCodeAt(0), 0);
  const baseTemp = 15 + (hash % 20);
  ...
  const conditions = ['sunny', 'cloudy', 'rainy'] as const;
  const condition = conditions[hash % 3];
  ...
}
```

```ts
// runtime.rn.tsx (around line 17)
function getMockWeather(location: string, units: 'metric' | 'imperial' = 'metric') {
  const hash = location.split('').reduce((acc, c) => acc + c.charCodeAt(0), 0);
  const baseTemp = 15 + (hash % 20);
  ...
  const conditions = ['☀️', '☁️', '🌧️'] as const;
  const condition = conditions[hash % 3];
  ...
}
```

The web version returns a semantic key (`'sunny' | 'cloudy' | 'rainy'`), consumed by a separate `WeatherIcon` component that maps the key to a glyph. The RN version instead returns the emoji glyph directly as `condition`, with no equivalent lookup step. So the two runtimes are not just duplicated, they've already drifted in what shape they hand back for the same field — anyone touching one copy (e.g. adding a fourth condition, or wiring in a real weather API later) has to remember to update the other copy consistently, and the type of `condition` is not even shared between them today.

## Impact

Currently cosmetic only, since both are placeholder mock data (this feature has no real weather API integration yet, by design, per its own doc comments). But it's a maintenance trap: the duplication has no shared source of truth, and the two copies have already diverged in representation, so a future change to one is likely to silently miss the other.

## Suggested fix

Factor the shared hashing/derivation logic (temperature, humidity, wind — everything except the platform-specific condition representation) into one function, e.g. in a shared module consumed by both runtimes, and let each platform map the shared numeric/condition-index result to its own presentation (semantic key for web + `WeatherIcon`, emoji glyph for RN) at the call site.

## How this was found

Noticed incidentally while doing an unrelated pass over these files, not investigated further — flagging for someone with more context on the intended web/RN code-sharing strategy for this package.

boomsi:
## Summary

Closes #147.

The weather feature's `getMockWeather` was duplicated verbatim between `runtime.web.tsx` and `runtime.rn.tsx`, with the `condition` field silently drifted between them — web returns a semantic key (`'sunny'|'cloudy'|'rainy'`) consumed by `WeatherIcon`; RN returns the emoji glyph directly. Per the issue's suggested fix, this factors the platform-agnostic derivation logic into a shared module and lets each runtime keep its own condition representation.

## Changes

- **New** `src/mock-weather.ts` — shared `getMockWeather(location, units)` returning `{ temp, unit, conditionIndex, humidity, wind, windUnit }`. Pure function, no platform imports (no React / DOM / RN), safe to consume from both runtimes. `condition` is intentionally **not** part of the shared contract — only its index is.
- `runtime.web.tsx` — drops the local `getMockWeather`; keeps its local `WEB_CONDITIONS = ['sunny','cloudy','rainy']` and maps `conditionIndex` → semantic key for `WeatherIcon`. The existing `|| icons.sunny` fallback is preserved.
- `runtime.rn.tsx` — drops the local `getMockWeather`; keeps its local `RN_CONDITIONS = ['☀️','☁️','🌧️']` and maps `conditionIndex` → emoji.

Each platform still owns its condition presentation; only the derivation logic (and the `hash % 3` index) is now single-sourced.

## Behavior

Zero behavior change. `hash`, `temp`, `humidity`, `wind`, `windUnit` are identical, and `conditionIndex = hash % 3` matches the previous `hash % 3` indexing on both sides, with condition-array order unchanged.

## Test plan

- [x] `bun --filter @supramark/feature-weather build` — tsc passes (exit 0)
- [x] `bun --filter @supramark/feature-weather test` — 5 pass / 0 fail (8 expect calls)
- [x] `bun run feature:lint weather` — 14/14 checks pass, score 100/100
- [ ] Web runtime visual check (`:::weather` container renders same SVG icon + temp/details) — needs host app
- [ ] RN runtime visual check (emoji + temp/details) — needs host app / device

### DIFF
--- a/packages/features/containers/weather/src/mock-weather.ts
+++ b/packages/features/containers/weather/src/mock-weather.ts
@@ -1,0 +1,59 @@
+/**
+ * Weather mock data derivation (shared layer)
+ *
+ * Contains only platform-agnostic derivation logic (temperature / humidity /
+ * wind / condition index). The concrete condition representation (web semantic
+ * key / RN emoji) is mapped by each runtime and is intentionally not coupled
+ * here — see WEB_CONDITIONS in runtime.web.tsx and RN_CONDITIONS in
+ * runtime.rn.tsx.
+ *
+ * A real app should replace this with a real weather API call.
+ *
+ * @packageDocumentation
+ */
+
+/**
+ * Return type of getMockWeather
+ */
+export interface MockWeather {
+  /** Temperature (already converted per units) */
+  temp: number;
+  /** Temperature unit label, e.g. '°C' / '°F' */
+  unit: string;
+  /**
+   * Condition index (0..2), derived from the location hash.
+   * Each runtime maps it to its own platform representation.
+   */
+  conditionIndex: number;
+  /** Humidity (%) */
+  humidity: number;
+  /** Wind speed (already converted per units) */
+  wind: number;
+  /** Wind unit label, e.g. 'km/h' / 'mph' */
+  windUnit: string;
+}
+
+/**
+ * Derive deterministic mock weather data from a location name.
+ *
+ * The same location always derives the same result (pure function, no side
+ * effects).
+ */
+export function getMockWeather(
+  location: string,
+  units: 'metric' | 'imperial' = 'metric',
+): MockWeather {
+  // Derive a pseudo-random temperature from the location name
+  const hash = location.split('').reduce((acc, c) => acc + c.charCodeAt(0), 0);
+  const baseTemp = 15 + (hash % 20);
+  const temp = units === 'imperial' ? Math.round(baseTemp * 1.8 + 32) : baseTemp;
+  const unit = units === 'imperial' ? '°F' : '°C';
+
+  const conditionIndex = hash % 3;
+
+  const humidity = 40 + (hash % 40);
+  const wind = 5 + (hash % 20);
+  const windUnit = units === 'imperial' ? 'mph' : 'km/h';
+
+  return { temp, unit, conditionIndex, humidity, wind, windUnit };
+}
--- a/packages/features/containers/weather/src/runtime.rn.tsx
+++ b/packages/features/containers/weather/src/runtime.rn.tsx
@@ -8,27 +8,17 @@
 
 import React from 'react';
 import { View, Text, StyleSheet } from 'react-native';
+import { getMockWeather } from './mock-weather.js';
 import type { ContainerRNRenderArgs } from '@supramark/core';
 import type { WeatherData } from './feature.js';
 
 /**
- * Mock weather data (a real app should call a weather API)
+ * condition index → RN emoji (platform-specific, not in shared layer)
+ *
+ * Shared derivation (temp/humidity/wind/conditionIndex) lives in
+ * mock-weather.ts; this only maps the condition for this platform.
  */
-function getMockWeather(location: string, units: 'metric' | 'imperial' = 'metric') {
-  const hash = location.split('').reduce((acc, c) => acc + c.charCodeAt(0), 0);
-  const baseTemp = 15 + (hash % 20);
-  const temp = units === 'imperial' ? Math.round(baseTemp * 1.8 + 32) : baseTemp;
-  const unit = units === 'imperial' ? '°F' : '°C';
-
-  const conditions = ['☀️', '☁️', '🌧️'] as const;
-  const condition = conditions[hash % 3];
-
-  const humidity = 40 + (hash % 40);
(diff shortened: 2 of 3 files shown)


kookyleo (approved):
## Review: approving — provably zero behaviour change on both platforms

The concern going in was that issue #147 says the two copies' condition representation differs, so unifying them would have to change one side. It does not: the PR extracts only the platform-agnostic derivation, returns a `conditionIndex`, and leaves each runtime's condition presentation at its own call site. Both condition tuples were moved verbatim.

**Differential probe** — the new shared `getMockWeather` against the old web and old RN implementations (pasted verbatim from history), over 30 locations x 2 unit systems:

```
cases                             : 60
numeric identical vs OLD-WEB      : 60/60   (temp, unit, humidity, wind, windUnit)
numeric identical vs OLD-RN       : 60/60
condition identical vs OLD-WEB    : 60/60   WEB_CONDITIONS[conditionIndex] vs old .condition
condition identical vs OLD-RN     : 60/60   RN_CONDITIONS[conditionIndex]  vs old .condition
conditionIndex observed           : 0,1,2   all three branches exercised
deterministic across repeat calls : true
```

Inputs included CJK, accented Latin, an apostrophe, an emoji, leading/trailing spaces, `'0'` and `''`. The derivation is byte-identical, not merely equivalent: same `charCodeAt` sum, same `15 + hash%20`, same `Math.round(baseTemp*1.8+32)`, same `40 + hash%40`, same `5 + hash%20`, same `hash%3`. The RN emoji tuple survived the move with its variation selectors intact (`U+2600 U+FE0F`, `U+2601 U+FE0F`, `U+1F327 U+FE0F`) — worth checking explicitly, since a mangled `U+1F327` would render as a different glyph with no test to catch it.

**Render-path probes** (nothing in the repo loads these runtimes today): web 12/12 — `WeatherIcon` receives the same `type` prop as before; RN 12/12 — the rendered `<Text>` contains the same emoji. Both cover all three condition branches.

**No silent fallthrough.** Exhaustive grep for `condition` in the package returns exactly two consumption sites, both updated (`runtime.web.tsx:173`, `runtime.rn.tsx:146`). There is no `switch` on the condition value anywhere, and `WeatherIcon`'s `icons[type] || icons.sunny` fallback is unchanged.

**No platform leak.** `mock-weather.ts` has zero imports — no react, no react-native, no DOM globals — confirmed by importing it standalone with react-native unavailable.

Tests: `feature-weather` 5 pass / 0 fail, `build` exit 0, `tsc --noEmit` exit 0, `feature:lint weather --strict` 14/14 checks score 100/100, eslint clean, `check-cjk` clean, `lint:types` total 0, `license:check` clean.

### Follow-ups (non-blocking)

- The shared module has no test coverage, and the repo's gate cannot catch that: `weatherFeature` is a `ContainerFeature`, which has no `testing.syntaxTests` field, so `feature:lint` still scores 100/100. A ~15-line golden-value test on `mock-weather.ts` would lock the invariant this refactor rests on — relevant for whoever wires in a real weather API later, which is the scenario #147 was written about.
- `conditionIndex: number` (`mock-weather.ts:27`) is weaker than the union it replaced; `0 | 1 | 2` with `(hash % 3) as 0 | 1 | 2` would restore unrepresentable-by-construction.
- The `% 3` and the two 3-element arrays now live in three files with nothing enforcing agreement. Add a fourth condition and bump the modulo without touching both arrays, and web degrades silently to always-sunny while RN renders an empty `<Text>`. Exporting `CONDITION_COUNT` from the shared module and typing both arrays `readonly [string, string, string]` is cheap hardening.

> Verification scope: element-level only. No iOS/Android simulator and no browser here, so painted pixels, on-device emoji fallback and real `StyleSheet` behaviour are unverified — which matches the PR's own honest, unticked visual-check boxes.

PR #157, opened 2026-07-31
test+fix(rn): drive renderer with real parser output; render raw/blockquote/thematic_break

### LINKED ISSUE #125: [renderers/rn] Tests never invoke the real parser, so parser/renderer shape mismatches are invisible to CI
## The gap

Every test under `packages/renderers/rn/__tests__/` hand-builds its AST as an object literal. None of them calls `parse()`. So the renderer is only ever tested against the shapes the test author *believed* the parser produces — a mismatch between real parser output and the assumed shape is invisible to CI by construction.

## This is not hypothetical

Found during the #105 review. `Supramark.tsx`'s `INLINE_NODE_TYPES` omits `raw`, and `raw-html` is in the **default** feature set of `supramark-markdown` (`crates/supramark-markdown/Cargo.toml:30`), so it is a common path, not an edge case.

Real parser output for `- a <span>b</span> c` is:

```
[text "a ", raw "<span>", text "b", raw "</span>", text " c"]
```

Which renders as:

| branch | result |
|---|---|
| base / #101 | `Text("• " "a " "b" " c")` — one line, tags dropped, text intact |
| #105 | `View(Text("• a ") View[pl:16] Text("b") View[pl:16] Text(" c"))` — **three lines plus two empty indented Views** |

Same for `- line1<br/>line2`. Every RN test stayed green through this, because none of them ever saw a `raw` node — the hand-built fixtures do not contain one. It was found only by building the parser CLI and diffing real output against the assumed shapes.

## Suggested shape

At minimum one test that goes through `parse()` end to end for a document exercising the node types the renderer switches on — inline HTML, hard breaks, loose and tight lists, nested lists, blockquotes, code blocks, tables. It does not need to assert a full tree; asserting that every node type the parser emits has a rendering branch (rather than falling through to `null`) would already have caught this.

Note the parser is a Rust binary, so this needs `cargo build -p supramark-markdown` in the test path or a checked-in golden AST regenerated by a script. A golden AST is the cheaper option and still catches shape drift, as long as regeneration is a documented step rather than hand-editing.

mybee:
## Summary

Closes #125.

The RN renderer tests never invoked the real parser — every fixture was a hand-built AST, so parser/renderer shape mismatches were invisible to CI. The default parser emits `raw` inline-HTML nodes that `INLINE_NODE_TYPES` omitted, so `- a <span>b</span> c` split into three `<Text>` lines plus two empty indented `<View>`s instead of one line.

### Test infrastructure
- New `parse-smoke.test.tsx` drives `<Supramark>` with the real Rust wasm parser's output (built in the test script, mirroring `@supramark/core`).
- Asserts every node type the parser emits has a rendering branch, locked via a `KNOWN_UNHANDLED` characterization set — any new unhandled type fails CI, and any gap that gets a branch must be removed from the set.
- Loads the wasm artifact by dist path because bun's `mock.module` registry is process-wide and `SupramarkCache.test.tsx` leaks a fake parser mock that would otherwise intercept `parse()`.
- Focused regression: `- a <span>b</span> c` renders as exactly one `• a b c` `<Text>`.
- Exports `BLOCK_NODE_TYPES` / `INLINE_NODE_TYPES` from `Supramark.tsx` so the test asserts against the renderer's own contract, not a hand-maintained copy.

### Rendering fixes (the three gaps the test surfaced)
- `raw` (inline HTML): drop the tag, keep the surrounding text in one `<Text>` flow. Added to `INLINE_NODE_TYPES` so `list_item` no longer routes it through the block path.
- `blockquote`: indented, left-bordered block rendering its children.
- `thematic_break`: a 1px divider `<View>`.

## Test plan
- [x] `bun test` in `packages/renderers/rn` → 95 pass / 0 fail (was 90 before the new tests).
- [x] `tsc -p tsconfig.json --noEmit` clean.
- [x] Mutation-checked the type-coverage assertion both ways (removed a type from `KNOWN_UNHANDLED` → fails; added a renderer branch without updating the set → fails).


kookyleo (approved):
Verified the updated head `6e9e37b2` after merging current `main`.

- RN package: 115 tests / 224 expectations passed
- real-parser smoke: 5/5 passed
- RN build, ESLint, and Prettier passed
- all 15 GitHub checks passed
- no unresolved review threads

The raw/blockquote/thematic-break rendering and real parser integration look correct.

Issue #158, opened 2026-08-01
[rn-selection] Reverse the interaction decision: draw the whole selection UI ourselves


kookyleo:
## Decision reversal

PR #84 recorded an "Interaction direction (decided)" in `packages/renderers/rn-selection/SELECTION_PLAN.md`: of the two candidate directions — **complete the vendored native command bridge** vs. **go fully self-drawn** — the command bridge won. A committed single-block range is pushed down to the vendored `SelectableRichText` via `selectRange`, and iOS/Android then paint the highlight, the grab handles and the **system edit menu**. Cross-block ranges stay on our own coordinator overlay.

**We are reversing that decision.** The selection layer becomes fully self-drawn: highlight, handles and the action bar are all rendered by `@supramark/rn-selection` in JS, identically on every platform. The native text view is demoted to what it is good at — laying out text and reporting metrics.

The trade being made explicitly: we give up some platform-native polish (see "What we lose" below) in exchange for **one behavior across platforms** and **full product control over the selection UI**, which is what issue #45 actually asks for.

---

### 1. The native path only ever serves the case we least need

Native text selection is per-text-view by construction. `planNativeSelection` therefore vetoes anything that spans two blocks, so the native path can only ever light up for a **single-block** selection. But the requirement in #45 is *"希望标题，正文，列表 可以连续选中"* — cross-block continuous selection. That case is structurally out of reach for the native primitive and is served by our own overlay regardless.

So today we carry the full cost of a second selection implementation, and it pays off only on the degenerate case.

### 2. One document, two different interactions

Long-press one paragraph → native highlight, native grab handles, system edit menu (Cut / Copy / Look Up / Translate…).
Drag across two paragraphs → flat block-level rect, **no handles, no menu at all**.

Nothing in the UI explains why. From a user's point of view the same gesture produces two unrelated products.

### 3. The seam is already generating bugs

Two selection representations with no shared owner have produced, in the first two weeks of the package's life:

- **#132** (fixed in #135) — duplicate highlight: the overlay had no idea the bridge had handed the block to native, so both painted. The fix required publishing the bridge's private "pushed" record as an external store and threading a `yieldNodeId` through `computeOverlayRects` → `SelectionContext` → `SelectionOverlay`. That whole mechanism exists purely to reconcile the two layers.
[...]

PR #135, opened 2026-07-28
fix(rn-selection): yield overlay for native-taken block to stop duplicate highlight

boomsi:
Fixes #132.

## Problem

On `@supramark/rn-selection` (after #84 merged), long-pressing a single text block painted two overlapping highlights: the vendored native selection (system grab handles + edit menu) AND a full-width block rect underneath from `SelectionOverlay`. Both fired on the same `commit()`. See #132 for the full analysis.

## Root cause

`SelectionOverlay` derived rects purely from `snapshot.units` with no knowledge that `nativeBridge` had already pushed that block to the native side via `selectRange`. The bridge's pushed record was closure-private, so the overlay had no way to know which block to skip.

## Fix

Expose the bridge's pushed block as an external store and have the overlay yield for it:

- `nativeBridge.ts` — `createNativeBridge` returns `{ unsubscribe, pushedStore }`; `pushedStore` publishes the pushed nodeId (or null) and notifies subscribers on every push/clear.
- `overlay.ts` — `computeOverlayRects` gains a `yieldNodeId` param; blocks matching it are skipped.
- `SelectionContext.ts` / `SelectionRoot.tsx` — the pushed store is threaded onto the context.
- `SelectionOverlay.tsx` — subscribes to the pushed store and passes `yieldNodeId` through.

**Behavior:**
- Single-block commit → bridge pushes native + publishes nodeId → overlay yields for that block → only the native selection (handles + menu) paints.
- Cross-block commit → `planNativeSelection` vetoes (second owner) → `yieldNodeId` is null → overlay paints every covered block (unchanged).

## Verification

- iOS sim (RN 0.81 New Arch), clean metro + Xcode build cache: long-press single block → only the native selection layer; duplicate rect gone. Cross-block `Select A..B` → merged coordinator overlay as designed.
- Tests: 173 pass / 0 fail (was 167; +6 — 3 overlay-yield, 3 pushedStore).
- `tsc --noEmit` clean (rn-selection package).

## Scope

Coordinator layer only (`packages/renderers/rn-selection/src/coordinator/` + tests). No vendored native changes; no `Podfile.lock` / `project.pbxproj` artifacts.

Follow-up for the separate "tap-outside not returning to idle" issue is tracked in #133.

Issue #132, opened 2026-07-28
Duplicate selection highlight: coordinator overlay does not yield for a native-taken block


boomsi:
## Observed

On `@supramark/rn-selection` (PR #84, at `a866dfbe`), long-pressing a single text block paints **two overlapping highlights**:

1. the vendored native selection — system grab handles + edit menu (correct, expected for single-block);
2. a full-width block rect underneath, covering even the whitespace past the text — from `SelectionOverlay`.

Both fire on the same `commit()`. The native layer comes from the command bridge (`nativeBridge.ts` → `handle.selectRange`); the overlay layer comes from `SelectionOverlay`.

## Root cause (from the code)

`SelectionOverlay.tsx:32` derives rects purely from `snapshot.units`:

```ts
const rects = computeOverlayRects(registry.getBlocks(), snapshot.units);
```

It has no knowledge that `nativeBridge` has already pushed that block to the native side. So for a single-block committed range, the overlay still draws underneath the native handles.

`nativeBridge.ts` kept its pushed record in a closure-private variable (`pushed`), with no way for the overlay to read which block the native side had taken over.

## Fix direction (drafted)

- `nativeBridge.ts` — `createNativeBridge` returns `{ unsubscribe, pushedStore }`; the pushed block is published as an external store (`getPushed` + `subscribe`).
- `overlay.ts` — `computeOverlayRects` takes a `yieldNodeId` param and skips that block.
- `SelectionContext.ts` / `SelectionRoot.tsx` / `SelectionOverlay.tsx` — the pushed store is threaded onto the context; the overlay subscribes and passes `yieldNodeId` through.

Single-block commit → bridge pushes native + publishes nodeId → overlay yields for that block → only the native selection paints. Cross-block → bridge vetoes (`planNativeSelection` returns null on a second owner) → `yieldNodeId` is null → overlay paints every covered block.

## Verification

- iOS sim (RN 0.81 New Arch), clean metro + Xcode build cache: long-press single block → only the native selection layer; duplicate rect gone. Cross-block `Select A..B` → merged coordinator overlay as designed.
- Tests: 173 pass / 0 fail (was 167; +6). `tsc --noEmit` exit 0.

## Scope

Coordinator layer only (`packages/renderers/rn-selection/src/coordinator/`). No vendored native changes.

kookyleo:
[...]

PR #159, opened 2026-08-01
feat(rn-selection): draw the whole selection UI ourselves

### LINKED ISSUE #133: Tapping outside clears the native selection but leaves the selection store in `selected`
## Observed

On `@supramark/rn-selection` (PR #84, at `a866dfbe`), after a single-block selection is committed (native handles + edit menu shown), **tapping outside** clears the native selection (handles/highlight gone) but the coordinator store stays in the `selected` phase. The status panel keeps showing `selected` until the `Clear` button is pressed.

## Root cause (from the code)

The native side clears its selection in several paths — `handleDismissTap` → `clearTextSelection` (`RCTSelectableRichTextView.mm:188-196`), `editMenuInteraction:didEndForConfiguration` (menu dismiss, `:474-478`), and `textViewDidChangeSelection` when the range collapses to empty (`:213-224`). None of these emit an event to JS:

- `clearTextSelection` only mutates native state (`selectedRange`, `editable`, `selectable`, `resignFirstResponder`) — it has **no event outlet**.
- The vendored component's only JS-facing events are `onTextLongPress` and `onMenuAction`; there is no `onSelectionChange` / deselection event.

So the coordinator store never learns the user dismissed the selection, and stays in `selected`.

## Fix direction

Add an `onSelectionChange` uplink event to the vendored native component, and have the coordinator clear the store when the native selection empties.

Vendored native (in `packages/renderers/rn-selection/native/selectable-rich-text/`):
1. codegen spec — add `onSelectionChange` + `bubbledEvents`.
2. C++ `EventEmitter` — add an `OnSelectionChange` struct.
3. iOS `.mm` — emit on the three empty-selection paths above; payload `{selectionStart, selectionEnd}` with `NSNotFound` for an empty selection.
4. TS `types.ts` / `SelectableRichText.tsx` — add and forward `onSelectionChange`.

Coordinator (`packages/renderers/rn-selection/src/coordinator/`):
5. `blockSink` or `nativeBridge` listens for `onSelectionChange`; an empty native selection calls `store.clear()`.

**Loop guard**: the vendored view already tracks `_commandSelectionActive` / `_isSettingCommandSelection` to distinguish command-driven selection changes from user-driven ones. `onSelectionChange` must NOT fire while those are set, otherwise the bridge's downlink `selectRange` would round-trip into a `store.clear()` and cancel the very selection it just pushed.

## Scope

Touches the vendored native component (codegen / C++ / iOS / TS types) AND the coordinator listener. Android side (`SelectableRichTextView.kt`) would need the symmetric change.

## Severity

Medium — no crash or data loss, but a user-visible state inconsistency after a common interaction (dismiss by tapping outside). Comparable to the should-fix items already addressed in #84.

kookyleo:
Implements #158.

Reverses the "Interaction direction (decided)" recorded in #84: instead of pushing a committed single-block range into the vendored native text view and letting the platform paint the highlight, the grab handles and the **system edit menu**, `@supramark/rn-selection` now draws all three itself — the same way on iOS and Android.

The reasoning is in #158; the short version is that native text selection is per-text-view, so the bridge could only ever light up for a single block while the actual requirement (#45) is cross-block continuous selection that our overlay had to draw regardless. Two owners for one selection produced #132 and #133, and the system menu left the action bar outside product control.

## What changed

**The text view becomes a metrics source.** `SelectableBlock` renders a plain React Native `<Text>` and publishes what `onTextLayout` reports — one entry per laid-out line, public RN API on both platforms, Paper and Fabric alike. No native selection component is on the render path any more.

New pure modules, all unit-tested:

- **`metrics.ts`** — the whole contract between a rendered block and the selection UI: a line table plus `offsetAtLocalPoint` (point → offset) and `rectsForRange` (range → rectangles). Carries an optional `charXs` slot so an exact per-character provider can drop in later without touching a consumer.
- **`coordinator/overlay.ts`** — `computeSelectionRects`: per-line highlight rectangles, projecting the document range onto each covered block through the existing `rangeToSegmentSelection`. A block with no metrics yet falls back to its whole rect, so nothing regresses while a provider is missing.
- **`coordinator/handles.ts`** + `SelectionHandles.tsx` — drag handles from the first and last rectangle; knobs sit clear of the line, and the grab radius is far larger than the drawn knob.
- **`coordinator/toolbar.ts`** + `SelectionToolbar.tsx` — the action bar. Placement arithmetic (prefer above, flip below, clamp into the viewport, arrow tracks the selection centre), host-supplied `toolbarItems`, and a `renderToolbar` escape hatch for hosts that want their own component.
- **`coordinator/gesture.ts`** — pure state machine `idle → pending → extending | handle → idle`, fed touch events plus an injected clock, so the long-press threshold is tested by advancing a number rather than waiting on one. Tap-to-dismiss lives here, which is the structural fix for #133.
- **`words.ts`** — word granularity for the long press: `Intl.Segmenter` where present, a script-class scan where it is not (the normal case on Hermes). Boundaries widen to grapheme clusters, so an emoji or a ZWJ sequence never selects in half.

**Retired with the bridge:** `nativeBridge.ts`, `blockSink.ts`, the overlay's `yieldNodeId`, `SelectionContext.nativePushed`, `nativePrimitive.ts`'s `TextSegmentHandle` command surface, the vendored-type shim, and the vendored peer dependency (also dropped from the example app, so its pod no longer autolinks). `native/selectable-rich-text` stays in the tree, unwired, as the natural home for the exact-metrics follow-up — upstream credit intact.

**Kept:** the whole model layer (linearize / resolve / serialize / tables / grapheme safety), the registry, hit-testing, the store, and the pure offset-mapping half of `segmentAdapter` — that mapping is exactly what projects a document range onto one block's text metrics.

## Consequences

- The package is plain TypeScript against public React Native APIs. No native module, no pod, no codegen spec, no autolinking. **The Fabric-only and Android >= 0.85 floors are gone**; the peer range returns to `react-native >= 0.72`.
- `SelectionCopyRequest` keeps its exact shape, so a host that already handles `onCopy` needs no change — only the origin of the action moved, from a `UIAction` in a system menu to a React button we own.
- The same pure geometry will serve `@supramark/web` once a `Range.getClientRects()` provider fills the same line table.

## Known limitations

- **Within-line position is interpolated.** `onTextLayout` reports lines, not characters, so a position inside a line is interpolated across the line's advance width: exact for monospaced text, off by up to about half a character in proportional text. Accurate enough to place a highlight edge and a handle; `charXs` exists for the exact provider, tracked as M-C in #158.
- **The magnifier is gone.** Deliberately not reimplemented. This is the real cost of the reversal and it is stated as such in `SELECTION_PLAN.md`.
- **Accessibility parity is not addressed here.** Leaving the platform selection layer means leaving its VoiceOver / TalkBack affordances; that is M-D in #158 and must not be treated as closed.
- **No drag autoscroll yet.**

## Verification

- **236 tests pass / 0 fail** (was 173): +59 covering metrics, word boundaries, handles, toolbar placement, the gesture machine, copy requests, registry metrics preservation, and the rewritten overlay.
- `tsc --noEmit` clean (package), `bun run lint:types` exit 0, `eslint --max-warnings 0` clean, `scripts/check-cjk.mjs` clean.
- **Not verified on device.** The touch dispatch and responder negotiation with an enclosing `ScrollView`, the long-press threshold and move tolerance, the handle knob geometry and hit slop, and whether `onTextLayout` line coordinates need adjusting for padding or `lineHeight` all need a real device. The decision-making those feed is unit-tested; the wiring is not, and cannot be here. That is M-B in #158.

## Closes / relates

- Implements #158
- Structurally resolves #133 (no native selection state left to fall out of sync) — proposing to close it once M-B confirms on device
- Removes the mechanism added by #135 for #132
- Serves the interaction half of #45; the renderer-side block-structure work described there is unchanged and still open


boomsi (commented):
@kookyleo

## Re-review: suite / types / lint clean — but on-device issues block merge

### Verified statically and by running the suite at HEAD (`27b367fe`)

Checked the PR out to a working tree and read every new module end to end:
`gesture.ts`, `hitTest.ts`, `handles.ts`, `metrics.ts`, `words.ts`,
`overlay.ts`, `toolbar.ts`, `SelectionRoot.tsx`, `SelectableBlock.tsx`,
`segmentAdapter.ts`, `registry.ts`, `SelectionContext.ts`.

- ✅ `bun --filter @supramark/rn-selection test` → **236 pass / 0 fail**, 588
  `expect()` calls, 16 files — matches the PR's claim (was 173).
- ✅ `tsc --noEmit -p packages/renderers/rn-selection/tsconfig.json` → clean.
- ✅ `eslint packages/renderers/rn-selection/src --max-warnings 0` → clean.
- ✅ `scripts/check-cjk.mjs` → clean.
- ✅ Vendored-peer removal holds: `@boomsi/react-native-selectable-text` is
  gone from `examples/react-native/package.json`; `Podfile` has no
  `selectable` reference so autolinking already drops it, and `ios/Pods/` is
  gitignored so `pod install` rebuilds without the vendored pod. (A stale
  `Pods/` dir still referenced it locally — clears on reinstall, not a source
  issue.)
- ✅ Peer range back to `react-native >= 0.72` (Fabric-only / Android >= 0.85
  floors gone) — confirmed in `package.json`.
- Added three demo blocks (multi-line English, ZWJ-emoji, CJK) to
  `SelectionDemo.tsx` to exercise cross-line / grapheme / CJK paths on device;
  kept ASCII-only via `\u` escapes (check-cjk still clean).

The model layer (linearize / resolve / serialize / grapheme safety), registry,
hit-testing, store, and the pure offset-mapping half of `segmentAdapter` are
sound and well-tested. The reversal off the native bridge to a self-drawn UI is
architecturally the right call: one owner for one selection, no per-text-view
native limit on cross-block ranges, identical iOS/Android paths.

### On-device findings (iOS only — Android NOT yet tested)

I could not run a device, so a reviewer ran the demo on iOS. These are real
observations, not speculation. Each is tagged with whether it falls inside the
PR's own "Known limitations" list (M-B = on-device wiring unverified, M-C =
exact-metrics follow-up, M-D = accessibility).

#### Hard — not in the PR's known-limitations list

- 🔴 **Toolbar covers the start handle.** `computeToolbarPlacement`
  (`toolbar.ts:125`) prefers above (`y = anchor.y - gap - size.h`, gap 8); the
  start handle knob sits at `first.y - R`, `R = 6` (`handles.ts:63`). For a
  single-line selection `anchor.y == first.y`, so the toolbar bottom
  (`first.y - 8`) lands inside the knob span `[first.y - 12, first.y]` — ~4pt
  overlap. Placement only avoids viewport margins, never the handle knob. Not
  called out anywhere in the PR; looks like a genuine gap.

#### Listed under M-B, but the root cause is pin-down-able from code (worse than "unverified")

- 🟡 **Handle drag → page scrolls (touch leaks to ScrollView).** Listed under
  M-B ("responder negotiation with an enclosing ScrollView"). On a short
  non-scrollable doc this read as "can't drag the handle at all"; once the doc
  overflowed the viewport it read as "handle drags *and* page scrolls." Root
  cause is concrete: `onStartShouldSetResponder` returns `gesture.isActive()`
  (`SelectionRoot.tsx:294`), which is necessarily `false` at touch start —
  handle detection happens in `gesture.touchStart`, which runs *after* the
  responder query — so the root never claims the responder on a handle grab;
  `ScrollView` does, and `onMoveShouldSetResponder` can only wrestle it back
  mid-drag. A start-point `handleAt` pre-check in `onStartShouldSetResponder`
  would fix both faces.

- 🟡 **Long-press fires inconsistently.** Listed under M-B ("long-press
  threshold"). Two modes seen: (a) selection appears while the finger is still
  down = `setTimeout(400ms)` fires `tick` on time; (b) selection appears only on
  finger-up = the timer was delayed by the JS/RN bridge and the `touchEnd`
  fallback (`gesture.ts:199`) honoured it. Separately, `moveTolerance = 10px`
  (`gesture.ts:43`) is tight enough that micro-jitter during a held press can
  `reset` to "it was a scroll", producing "finger-up doesn't select either."

#### Listed under M-B (`onTextLayout` coords / timing)

- 🟡 **Row-1 long-press highlighted row 3 (≈2-row skew).** Listed under M-B
  ("`onTextLayout` line coordinates need adjusting for padding or
  `lineHeight`"), but a 2-row skew is larger than padding/lineHeight drift.
  It disappeared after reload and once the doc grew — pointing at a timing
  race rather than a deterministic mapping bug. One concrete suspect from code:
  `originRef` (`SelectionRoot.tsx:251`) is set once in `onLayout` via
  `measureInWindow` and **not refreshed on scroll**, so a touch after scrolling
  uses a stale window origin and shifts every coordinate.

- 🟡 **Highlight / handle sit above the characters.** Same line-coordinate
  root: highlight `y = line.y` (`metrics.ts:236`) and the handle knob derives
  from the highlight rect (`handles.ts:63`), so a `line.y` semantic mismatch
  drags both up. Falls under M-B ("handle geometry / hit slop" + onTextLayout).

- 🟡 **Blank-space long-press: toolbar flash → hide → reappear, ends on the
  word to the right.** Listed under M-B (metrics timing). The flash/hide/reappear
  lines up with the toolbar visibility gate
  `snapshot.phase === 'selected' && rects.length > 0` (`SelectionToolbar.tsx:49`)
  flickering as `computeSelectionRects` returns `[]` for a frame while metrics
  settle, then re-populates; "ends on the word to the right" is the within-line
  interpolation landing the offset in the adjacent word.

#### Listed under M-C (within-line interpolation) — but CJK behaviour is sharper than "half a character"

- 🟡 **CJK: highlight stops mid-character; dragging a handle snaps a whole
  row.** Listed under M-C ("within-line position is interpolated … off by up to
  about half a character"). But CJK is `single`-class (`words.ts`) and on a
  fixed-width run `xForOffset`'s linear interpolation (`metrics.ts:151`,
  `line.x + line.w*k/span`) is *exact* — a half-character drift on pure Han
  text suggests `line.w` / `span` don't actually correspond on device (e.g.
  `onTextLayout` `width` carrying trailing slack), which is a sharper claim than
  "approximate." Handle-drag jumping a whole row on a small move is the same
  root: offset mapping on that block is unstable. Needs an instrumented run to
  pin — `line.y/h/w`, `rect`, `contentOffset`, `origin`, `offsetAtLocalPoint`
  in/out.

- 🟡 **Latin long-press sometimes selects only part of a word.** M-C
  interpolation category; on proportional Latin this is the documented
  half-character case, though it may also be carrying the same offset-system
  mismatch as the CJK item above.

#### Deliberately not raised

The PR already declares these, so no action is asked: magnifier gone
(deliberate), no drag autoscroll, accessibility parity (M-D), not verified on
device (M-B).

### Verdict

The pure layer is in good shape and the suite / types / lint gate is green, so
the **direction** is not in question. But on-device behaviour is not yet
shippable: the toolbar/handle overlap is a deterministic geometry bug (not in
the known-limitations list), and the responder issue makes handle dragging
actively fight the `ScrollView` on any scrollable document. Both need code
changes plus another iOS pass — and an Android pass, which hasn't happened.
**Not ready to merge as-is.**

boomsi (commented):
@kookyleo

## Re-review at `f5278583`: responder/toolbar direction is right — but `toRootSpace` reintroduces a coordinate-space bug

### Verified at `f5278583`

- ✅ `bun --filter @supramark/rn-selection test` → **241 pass / 0 fail**, 595
  `expect()` calls, 17 files (was 236/16; +5 for responder + toolbar placement).
- ✅ `tsc --noEmit -p packages/renderers/rn-selection/tsconfig.json` → clean.
- ✅ `eslint … --max-warnings 0`, `scripts/check-cjk.mjs` → clean.

### The fix lands the direction, but one line undoes it

`onStartShouldSetResponder` now does `handleAt(toRootSpace(e))`
(`SelectionRoot.tsx:271`) — exactly the start-point pre-check the first review
asked for, and `avoidRects` in `toolbar.ts` is the right idea for the knob
overlap. But `toRootSpace` (`SelectionRoot.tsx:213-219`) now prefers
`locationX/locationY`:

```ts
const { locationX, locationY, pageX, pageY } = e.nativeEvent;
if (Number.isFinite(locationX) && Number.isFinite(locationY)) {
  return { x: locationX, y: locationY };          // ← child-space, not root
}
return { x: pageX - originRef.current.x, y: pageY - originRef.current.y };
```

`locationX/locationY` is the touch position relative to the **target View that
received the touch** — the deepest child, e.g. the `<Text>` inside a
`SelectableBlock` — **not** relative to `SelectionRoot`. Everything downstream
(`handleAt` / `hitTestHandle`, `chooseBlock`, `block.rect`, the overlay rects)
is in `SelectionRoot` space, so that branch returns a point in the wrong
coordinate space. On a real device this shows up as two distinct failures:

- 🟡 **Handle drag still scrolls the page (#5 not actually fixed).**
  `onStartShouldSetResponder` → `handleAt(toRootSpace(e))` tests a child-space
  point against root-space knob geometry → misses → returns false → the root
  never claims the responder → `ScrollView` keeps it. On a short,
  non-scrollable doc this still reads as "can't drag the handle"; once the doc
  overflows the viewport it reads as "handle drags *and* page scrolls."
- 🔴 **New regression: long-press position and selection don't match.**
  `onTouchStart` → `gesture.touchStart(toRootSpace(e))` (`SelectionRoot.tsx:227`)
  stores a child-space `pressPoint`; 400 ms later `wordAt(pressPoint)` runs
  `chooseBlock` against root-space `block.rect`s → resolves the wrong
  block/offset → the highlight lands somewhere else. This was not reported
  before this commit; the old `pageX - origin` path at least stayed in root
  space. (The prior "#1 row-1 → row-3 skew" was an origin-staleness symptom on
  the old code; this commit compounds it into a full mismatch.)

**Fix is one line:** drop the `locationX/Y` branch and keep
`pageX/pageY - originRef`. `originRef` is now refreshed on every touch via
`refreshRootOrigin`, and although `measureInWindow` is async (so the first
event of a gesture may use a slightly stale origin), that is a far smaller and
transient error than a child-space point that is wrong on every event.

### Still open from the first review (unaddressed by this commit)

- 🟡 **#1 row skew** — still present; was origin-staleness before, now
  compounded by the `locationX/Y` regression above.
- 🟡 **#4 toolbar still partially occludes the handle.** `avoidRects` fixes the
  above/below main paths, but the neither-fits clamp branch
  (`y = clamp(above, margin, viewport.h - margin - size.h)`) does not keep
  avoiding the knob once it clamps into the viewport — so when the selection is
  near the top and the bar can't fit above, it clamps to `margin` and can land
  back on the knob.
- 🟡 **#2 handles / highlight sit above the characters** — `metrics.ts`
  `line.y`/`line.h` semantics, untouched.
- 🟡 **#3 / #7 / #10 CJK half-character highlight, whole-row snap on handle
  drag** — `xForOffset` interpolation / `line.w`·`span` mismatch on device,
  untouched. (CJK is fixed-width, so the linear interpolation is *expected* to
  be exact here — the half-character drift on pure Han suggests the reported
  `line.width` doesn't correspond to `span` characters.)
- 🟡 **#6 blank-space long-press: toolbar flash → hide → reappear** — metrics
  timing, untouched.
- 🟡 **#8 long-press fires inconsistently** — `gesture.ts` `setTimeout(400ms)`
  / `moveTolerance = 10px`, untouched.

### Verdict

The responder / origin-refresh / toolbar-avoidance direction is correct; the
`locationX/Y` line is the single change undoing the responder fix and adding a
new regression. Fix that one line and re-test handle drag + long-press
positioning on device (plus an Android pass, still not done). **Still not
ready to merge.**

boomsi (commented):
@kookyleo

## Re-review at `0663526f`: coordinate / hit-test / toolbar fixed — remaining on-device issues

### Verified at `0663526f`

- ✅ `bun --filter @supramark/rn-selection test` → **244 pass / 0 fail**, 602
  `expect()` calls, 17 files.
- ✅ `tsc --noEmit -p packages/renderers/rn-selection/tsconfig.json` → clean.
- On-device pass on iOS (Android still not tested).

### Fixed — confirmed on device

- ✅ **Long-press position now matches the finger.** The `toRootSpace`
  `locationX/Y` regression is gone (`7c3955d5`); pressing a word selects that
  word.
- ✅ **#6 blank-space long-press no longer selects a neighbour.**
  `containingBlock` (`e14cc944`) returns null off-block, so a press in the gap
  between blocks does nothing instead of grabbing the nearest one.
- ✅ **#4 toolbar no longer covers the handle.** `avoidRects` +
  `overlapArea` side selection (`7c3955d5`) + `HANDLE_TOUCH_RADIUS`
  (`e14cc944`) hold across viewport-top / viewport-bottom edge cases.
- ✅ **CJK selects per character; #10 whole-row snap on handle drag is gone.**

### Remaining (all package-side, not the example)

1. 🟡 **#5 + new: handle drag still scrolls the page; first drag doesn't move,
   second does; the other handle drifts.**
   - Suspected cause: iOS `UIScrollView`'s native `panGestureRecognizer`
     pre-empts the RN responder — `onResponderTerminationRequest` cannot veto
     it, so `onResponderTerminate` fires → `gesture.cancel()` aborts the first
     drag; the second works because residual state differs. The "other handle
     drifts" points at `fixedPoint` capture/retention in `gesture.ts` handle
     state (`beginAt(fixedPoint)` not actually pinning the opposite edge).
2. 🟡 **#2 / #3 / #7: half-character selection + highlight/handle position
   drift, on both Latin and CJK.**
   - Suspected cause: `xForOffset` linear interpolation (`metrics.ts:151`)
     without `charXs`. On CJK (fixed-width) the interpolation is *expected* to
     be exact, so the drift there suggests `onTextLayout`'s `line.width` does
     not correspond to `span` characters (e.g. trailing slack in the reported
     width).
3. 🟡 **#8 + new: long-press does not fire while the finger is down; it only
   appears on finger-up.**
   - Suspected cause: `setTimeout(400ms)` is delayed by the RN bridge and the
     `touchEnd` fallback (`gesture.ts:199`) honours it instead.
     `moveTolerance = 10px` may also `reset` on micro-jitter during the hold,
     producing "finger-up doesn't select either".
4. 🟡 **New: long-press → drag, the toolbar flashes then hides, reappears on
   release.**
   - `commit()` shows the bar, `extending` → `extendTo` hides it (`selecting`),
     release `commit` shows it again — this sequence is by design
     (`gesture.ts` `beginWordSelection` comment). The *flash* is acute when
     `tick` fires late (during the drag) so `commit` and `extendTo` land
     back-to-back; shares the #8 timer root.
5. 🟡 **New: the `Select A..B` / `Select in block` buttons have no effect.**
   - Suspected cause: `gesture.touchEnd` in the `pending` state calls
     `store.clear()` (tap-to-dismiss). A button tap also enters `pending`
     (`gesture.touchStart` does not distinguish text from button touches), so
     the selection the button just set is cleared on release. `store` itself is
     fine — `beginAt`/`extendTo`/`commit` work; the clear undoes them.
6. 🟡 **New: long-press inside the FlatList does nothing.**
   - Cause: `SelectableBlock.onBlockLayout` (`SelectableBlock.tsx:140`) uses
     `e.nativeEvent.layout.x/y`, which is **parent-relative**, and the block
     assumes it is a direct child of `SelectionRoot`. Inside a FlatList cell
     the parent is the cell, so `block.rect` is stored in cell space but
     consumed as root space → `containingBlock` never hits. This is a
     coordinate-space assumption in the package, not the demo.

### Verdict

Three commits moved the needle on the coordinate / hit-test / toolbar cluster,
and four of the original findings are closed. What remains is three deeper
buckets: **responder (#5 / new-1 — iOS native gesture pre-emption)**,
**timing (#8 / new-3 / new-4 — `setTimeout` unreliability)**, and
**coordinate assumptions (#2/#3/#7 interpolation + new-6 FlatList nesting +
new-5 gesture clearing button selections)**. **Still not ready to merge.**

boomsi (commented):
@kookyleo

## Re-review at `8a424c4c`: most on-device issues fixed — remaining FlatList scroll + coordinate drift (iOS + Android)

### Verified at `8a424c4c`
- ✅ `bun --filter @supramark/rn-selection test` → **262 pass / 0 fail**, 635 `expect()` calls.
- ✅ `tsc --noEmit` clean.
- On-device: iOS **and** Android.

### Fixed — confirmed on both platforms
- ✅ `#6` blank-space long-press no longer erratic / no menu flicker.
- ✅ `#8` / new-3: long-press fires while the finger is down (no longer waits for release).
- ✅ new-4: menu no longer flashes on long-press → drag.
- ✅ `#4`: toolbar does not cover the handle.
- ✅ `#10`: CJK + Latin per-character selection works.
- ✅ Android: handles can be dragged (was completely stuck).

### Remaining

**FlatList nesting — `measureLayout` fix incomplete + PanResponder blocks nested scroll:**

1. 🔴 **FlatList cannot scroll on iOS; intermittent on Android.** `rootPanResponder`'s `onStartShouldSetResponder: () => enabled` claims *every* touch — including ones meant for a nested FlatList's scroll. On touchMove past tolerance the responder is released to the *parent* ScrollView, not the FlatList, so the page scrolls instead. Suggested fix: claim only when `containingBlock(point) !== null || handleAt(point) !== null`.

2. 🔴 **After scrolling the FlatList, selection/menu position drifts (both platforms).** `onBlockLayout`'s `measureLayout` only fires on layout — **scrolling does not re-fire it**, so `block.rect` stays at the pre-scroll position. Observed: long-press lands between two lines (iOS), text sticks to the top edge of the selection (Android), and after selecting then scrolling the FlatList the highlight + menu stay at the old screen position while the text scrolls away. Needs an onScroll-driven re-measure.

**Responder:**

3. 🟡 **iOS: dragging a handle still scrolls the page.** Android handles work now, but iOS `UIScrollView`'s native `panGestureRecognizer` still pre-empts the handle `PanResponder` despite `onShouldBlockNativeResponder: () => true`.

4. 🟡 **Android: page / FlatList scroll intermittently conflict** — sometimes the page scrolls, sometimes neither does. Same responder root as item 1.

5. 🟡 **Dragging one handle moves the other too.** `gesture.ts` handle state: `fixedPoint` is captured on grab, but `beginAt(fixedPoint)` on each move does not actually pin the opposite edge — it drifts.

**Coordinate precision:**

6. 🟡 **Long-press position skews right by ~one character** (both platforms, most visible on CJK / large font — press the 3rd char, select the 4th). Systematic offset, not an interpolation edge.

7. 🟡 **Half-character selection still present** (both platforms, Latin + CJK). `xForOffset` interpolation / `line.w`·`span` mismatch (earlier `#2` / `#3` / `#7`).

**Toolbar lifecycle:**

8. 🟡 **Tapping a toolbar action does not dismiss the selection or the menu.** `runToolbarItem` fires `onCopy` but never `clear`s the store. Native edit menus dismiss on action — is that intended here?

### Non-issue (clarifying the demo buttons)
`Select A..B` selects across many blocks because it spans `h → zh1` (cross-block). `Select in block` selects "two words + half a word" because its offset 0→10 is **UTF-16 character-level**, not word-boundary-snapped — offset 10 lands inside "the". Both are expected for programmatic selection.

### Verdict
The responder / toolbar / long-press cluster is largely resolved and both platforms are usable for static blocks. The blockers are the **FlatList nesting pair** (scroll blocked + coordinate drift after scroll) — both stem from the `8a424c4c` PanResponder + `measureLayout` changes — plus the iOS native-gesture handle issue and the coordinate-precision items. **Not ready to merge** until the FlatList pair is addressed; the rest could land as known limitations.

boomsi (commented):
@kookyleo

## Re-review at `217a1e5e`: FlatList scroll still broken + new out-of-viewport selection (iOS + Android)

### Verified at `217a1e5e`
- ✅ `bun --filter @supramark/rn-selection test` → **268 pass / 0 fail**, 650 `expect()` calls, 17 files.
- ✅ `tsc --noEmit` clean.
- On-device: iOS **and** Android.

### Inferred fixed this round (not reported again)
- ✅ Long-press skew by one character (`offsetInsideLineX` floor) — only "half-character highlight, a few px" remains (see remaining 4); "press 3rd, select 4th" no longer reported.
- ✅ Toolbar action does not dismiss (`runToolbarItem` + `store.clear()`) — not reported again.
- 🟡 Selection position after scroll (`refreshLayouts`) — Android eventually catches up but with ~500ms lag (see new 7); iOS FlatList cannot scroll, unverified.

### Remaining (217 did not fix)

1. 🔴 **FlatList scroll still broken (iOS + Android)**
   - iOS: swiping on the FlatList scrolls the whole page (outer ScrollView); after selecting text it sometimes scrolls.
   - Android: FlatList / page scroll alternate — after scrolling the FlatList you cannot immediately scroll the page, and vice-versa; sometimes neither scrolls.
   - 217 changed `onStartShouldSetResponder` to claim only on `handleAt || containingBlock` hit, but nested FlatList scrolling still does not work. Suspected: the outer ScrollView's native gesture priority still pre-empts the FlatList's internal scroll.

2. 🟡 **iOS: handle drag still scrolls the page (+ selection changes with scroll)**
   - Dragging a handle scrolls the page and the selection range changes too (not just the page moving — the selection is being modified).
   - Suspected: `UIScrollView`'s native `panGestureRecognizer` pre-empts the touch, `onShouldBlockNativeResponder` does not block it; selection changing with scroll means the drag's `pointAt` resolves to wrong coordinates after scroll.

3. 🟡 **Dragging one handle moves the other (iOS + Android)**
   - `gesture.ts` handle state `fixedPoint` does not pin the opposite edge.

4. 🟡 **Half-character highlight (iOS + Android)**
   - Long-press word selection: the highlight is off by a few px (sometimes left, sometimes right), but the actual selected character is complete.
   - Root cause: `rectsForRange` uses `xForOffset` linear interpolation for the **highlight boundary**, no `charXs`. Note `offsetInsideLineX` fixed the **hit** (which character), not the **highlight boundary** (where it is drawn) — two distinct issues.

### New (found this round)

5. 🔴 **FlatList out-of-viewport selection (iOS + Android) — package bug, not the example**
   - After selecting inside a fixed-height FlatList, dragging a handle can drag past the height limit and keep selecting invisible FlatList items; the highlight is drawn in root space, on top of other elements. Long-pressing at the outer edge of the viewport (just outside the visible area) also selects invisible text.
   - Verdict: **package bug**. The overlay is an `absoluteFill` child of `SelectionRoot` and does not respect FlatList viewport clipping; `measureLayout` measures FlatList items into root space, so handle-drag `pointAt` hits any block (including off-screen ones) and the highlight is drawn outside the viewport. The example's fixed-height nesting exposes it, but the root cause is in the package (overlay/selection not clipped to the FlatList visible area, handle drag not clamped to visible blocks).

6. 🟡 **Android: handle drag inside FlatList is treated as scroll (selection unchanged)**
   - Dragging a handle on the page does not scroll (OK), but dragging a handle inside the FlatList is recognized as FlatList scroll and the selection does not change.
   - Suspected: the FlatList's ScrollView grabs the handle touch, the handle PanResponder does not get it.

7. 🟡 **Android: after selecting in FlatList then scrolling, selection/menu lags ~500ms before following**
   - Scrolling the FlatList: the selected text scrolls first, then the selection style + menu follow (about 500ms lag).
   - Suspected: `refreshLayouts`' `measureLayout` is async + `refreshSelectionLayoutsSoon`'s rAF/setTimeout triple-scheduling causes the selection update to lag behind the text scroll.

### Verdict
217 fixed the hit-skew and toolbar dismiss, but FlatList scroll (remaining 1) is still broken, and this round exposed out-of-viewport selection (new 5, package bug) and the Android handle being grabbed by FlatList (new 6). **Not ready to merge** — the FlatList nesting pair (scroll + out-of-viewport) is the main blocker.

boomsi (commented):
@kookyleo

## Re-review at `fd949421`: SelectionViewport fixes scroll + iOS handle-scroll — remaining handle-loss + cross-viewport selection

### Verified at `fd949421`
- ✅ `bun --filter @supramark/rn-selection test` → **283 pass / 0 fail**, 687 `expect()` calls, 17 files.
- ✅ `tsc --noEmit` clean.
- On-device: iOS **and** Android.

### Fixed
- ✅ FlatList scroll — `SelectionViewport` wraps the FlatList, it scrolls correctly; page scroll is also normal.
- ✅ iOS: handle drag no longer scrolls the page, and the selection no longer changes with scroll (`onGestureActiveChange` + `scrollEnabled`).
- 🟡 500ms lag → ~<100ms jitter (sync offset helps, but a small async remains).

### Remaining (still present, both platforms unless noted)

1. 🟡 **fixedPoint**: dragging one handle still moves the other. `gesture.ts` has not pinned the opposite edge.
2. 🟡 **Half-character highlight**: highlight still off by a few px (the actual selected char is complete). `xForOffset` interpolation, no `charXs`.
3. 🟡 **Cross-viewport selection**: dragging a handle outside the FlatList still selects text outside the FlatList. (The earlier "highlight covers other elements" is gone, but the selection range still crosses the viewport boundary.)
4. 🟡 **Android: handle still grabbed by FlatList** — intermittent reproduction.

### New (this round, both platforms unless noted)

5. 🔴 **Handle lost after dragging outside the FlatList** — dragging the right handle past the FlatList edge, the selection stops mid second-to-last row and the right handle disappears.
6. 🔴 **Handles not visible when selecting the leftmost 1-2 chars** in a FlatList — both handles invisible, sometimes only the left one.
7. 🔴 **Handle lost after vertical handle drag inside the FlatList** (not crossing the edge, not swapping sides) — on release the dragged handle can disappear.
8. 🟡 **During FlatList momentum scroll, elements outside the FlatList are unresponsive** (iOS + Android). Question: is this intended (SelectionViewport pauses interaction while momentum scrolling)?
9. 🟡 **Grey background on FlatList item press (iOS only; Android does not show it).** Not the example's style (`flatRow` / `flatList` have no `backgroundColor`); appears to be an iOS touch-visual side effect of the `SelectionViewport` / rootPanResponder touch handling. Needs confirmation.
10. 🟡 **~<100ms jitter on selection/menu when scrolling the FlatList after selecting** — the 500ms lag is gone but a small async remains.

### Questions for the author
- **Cross-viewport selection (3)**: cross-block selection is the PR's core design, but should dragging a handle past the FlatList viewport boundary be clamped to the viewport? Currently it selects text outside the FlatList.
- **Momentum-scroll pause (8)**: is pausing all interaction outside the FlatList during momentum scroll intended, or should it be scoped narrower?
- **Grey background (9)**: is the iOS press highlight expected, and should the package suppress it?

### Verdict
`SelectionViewport` resolved the FlatList scroll + iOS handle-scroll pair (the main blockers from the last review). What remains is the **handle-loss cluster (5/6/7, new)** — handle drag inside a FlatList is currently unreliable — plus **fixedPoint (1, historical)** and **cross-viewport selection (3, needs a clamp decision)**. **Not ready to merge** until the handle-loss cluster is addressed.

boomsi (approved):

PR #160, opened 2026-08-03
fix(rn): render-cache follow-ups from #92 review

### LINKED ISSUE #124: [renderers/rn] Render-cache follow-ups from the #92 review: unbounded by bytes, no clear API, policy-key leak
Carried over from the review on #92, which was merged with these open. Each was measured, not predicted. Ordered roughly by how likely it is to bite a real host.

### 1. The cache is default-on and bounded by entry count only

`packages/renderers/rn/src/renderCache.ts:37` + `packages/core/src/feature.ts` — `createConfigFromRegistry()` returns `options.cache: true`, and #92 is what makes that flag load-bearing. `sizeCalculator: () => 1` makes `totalSize` meaningless, so the bound is "100 entries" regardless of payload size.

A chat host scrolling 500 mermaid messages keeps the 100 most recent normalised SVGs — routinely 50–500 KB each — plus 100 parsed ASTs resident for the process lifetime, surviving every unmount.

Suggest a lower document default and/or a byte-aware `sizeCalculator` with a total-bytes cap.

### 2. No public way to drop the cache

`packages/renderers/rn/src/index.ts` — `clearReactNativeRendererCaches` is marked `@internal` and `renderCache.ts` is not re-exported from the package index, so a host reacting to an `AppState` memory warning has no supported API.

One line: `export { clearReactNativeRendererCaches as clearSupramarkRenderCache } from './renderCache';`

### 3. Policy-keyed caches are stranded forever

`renderCache.ts:91, 142` — `rendererCaches` is keyed `namespace:maxSize:ttl`, so every distinct policy strands a whole `AsyncRendererCache`. The comment at `:84-90` argues policy is static, but `resolveDocumentCachePolicy` (`Supramark.tsx:97-101`) *derives* `maxSize` via `Math.min(...)` over enabled engine policies — so a host enabling different engine subsets per message produces new policy keys without changing any code.

Verified: after 500 distinct policies the `maxSize:1` cache is still alive and still holding its entry.

Suggest keying by `namespace` alone and treating policy as a reconfiguration.

### 4. The first concurrent caller's `shouldRetain` silently overrides everyone else's

`renderCache.ts:47-77` — two rows over the same markdown, one with `options.cache:false` and one with `true`, resolve to the same policy and document key, so the second dedups onto the first's in-flight promise and inherits its retention.

Verified: `p1 === p2` is true and `cache.get('shared')` is `undefined` even though caller 2 passed `() => true`.

Suggest OR-ing the predicates across joiners, or folding retention intent into the key.

### 5. The doc comment describes an RN/Web parity that does not exist

`packages/core/src/feature.ts:1599-1604` says the cache applies to `@supramark/rn` / `@supramark/web`, but `packages/renderers/web/src/` has no document or diagram cache at all — `Supramark.tsx` only forwards `engineConfig.cache` into engine options. #92 creates an RN/Web divergence and documents the opposite.

Suggest saying RN-only and noting the divergence.

### Nits, same review

- `Supramark.tsx:48-52` — the cached `ParsedDocument.root` is a shared mutable AST and `readonly` is shallow and type-only. Verified: three mounts share one `root`; mutating a text node through the tree the first mount received changes what the third renders. A host `containerRenderers` annotating nodes in place would cross-contaminate every row. Suggest a dev-mode deep freeze on the snapshot.
- `renderCache.ts:216` — functions and Symbols still collide in `stableSerialize` (two distinct closures with identical source; two `Symbol('theme')`), and symbol-keyed props are dropped.
- `renderCache.ts:194, 206` — `new Set(seen)` per node makes the serializer O(depth²); a 2000-level chain takes 433 ms. Config objects are shallow so it does not bite today.
- `renderCache.ts:173-181` — structurally-equal `Date`/`Map`/`Set` defeat the cache: three mounts with freshly-constructed but equal-valued Dates produce 3 engine calls, and each miss inserts a junk entry that evicts a good one.

### Note for #91

The fix for #91 is opt-in — `renderCache.ts:98` defaults `enabled` to `false`. A host that hand-builds `config` rather than using `createConfigFromRegistry` will still hit #91.

mybee:
Closes #124.

Carries over every measured follow-up from the #92 review.

## Summary

- **Bound by bytes, not just count** — `LRUCache` gains a `maxBytes` cap; eviction respects both entry count and total estimated size. RN render cache lowers the default to `maxSize: 50` + `maxBytes: 8 MB` and uses a real per-namespace `sizeCalculator` (`svg.length` for diagrams, an AST walk for parsed docs) instead of `() => 1`, which made `totalSize` meaningless.
- **No more stranded caches** — `rendererCaches` is keyed by namespace alone; a policy change `reconfigure()`s the existing `AsyncRendererCache` in place instead of leaking a whole cache per distinct `maxSize:ttl` key (`resolveDocumentCachePolicy` derives `maxSize` from the enabled engine subset, so every subset change stranded a cache before).
- **`shouldRetain` no longer silently overridden** — the predicate is OR-joined across every caller that dedups onto the same in-flight promise, so a `() => true` joiner no longer inherits a first caller's `() => false`.
- **Public drop-cache API** — `export { clearSupramarkRenderCache }` so a host reacting to an `AppState` memory warning can release cached SVGs/ASTs.
- **`stableSerialize` tightened** — `Date`/`RegExp`/`Map`/`Set` compared by value (equal-valued distinct instances now hit the same key instead of missing and evicting a good entry); functions/symbols keep identity (no collision on identical source/description); symbol-keyed props no longer dropped; backtracked `seen`-set makes it O(depth), not O(depth²).
- **Dev deep-freeze of the shared cached AST** — catches a `containerRenderers` annotating nodes in place and cross-contaminating sibling rows; production keeps the freeze off the render path.
- **Doc fix** — `createConfigFromRegistry`'s comment no longer claims RN/Web cache parity; the runtime cache is RN-only (`@supramark/web` forwards `engineConfig.cache` but keeps no document/diagram cache).

## Test plan

- [x] `packages/core/__tests__/cache.test.ts` — byte-cap eviction, single-oversized-entry retention, `reconfigure` (maxSize / maxBytes / ttl)
- [x] `packages/renderers/rn/__tests__/renderCache.test.ts` — equal-valued Date/Map/Set/RegExp hit the same key; functions/symbols don't collide; symbol-keyed props not dropped; deep-chain O(depth); `shouldRetain` OR-join; namespace-only key (reconfigure, not strand); `clearSupramarkRenderCache` alias
- [x] `packages/renderers/rn/__tests__/SupramarkCache.test.tsx` + `DiagramNode.test.tsx` unchanged behavior
- [x] `bun run test` — all workspaces 0 fail
- [x] `eslint` on touched files + `lint:cjk` clean

### DIFF
--- a/packages/core/src/cache.ts
+++ b/packages/core/src/cache.ts
@@ -18,9 +18,24 @@ export interface LRUCacheOptions {
 
   /**
    * Optional value size estimator (used to estimate memory footprint).
+   *
+   * When {@link maxBytes} is set, entries are also evicted once total estimated
+   * size exceeds it, so this should return a byte-like number for the bound to
+   * be meaningful. Returning 1 (as the RN renderer did before #124) makes the
+   * byte cap a no-op.
    * @default (value) => JSON.stringify(value).length
    */
   sizeCalculator?: (value: unknown) => number;
+
+  /**
+   * Optional total-size cap. When set, `set()` evicts oldest entries until
+   * `totalSize <= maxBytes` (in addition to the entry-count `maxSize` bound).
+   * A single entry larger than `maxBytes` is retained alone (it cannot be
+   * evicted to make room for itself), so size is best-effort, not a hard
+   * guarantee. `undefined` disables the byte bound.
+   * @default undefined
+   */
+  maxBytes?: number;
 }
 
 interface CacheEntry<T> {
@@ -51,8 +66,9 @@ interface CacheEntry<T> {
  */
 export class LRUCache<T> {
   private cache: Map<string, CacheEntry<T>>;
-  private readonly maxSize: number;
-  private readonly ttl: number | undefined;
+  private maxSize: number;
+  private ttl: number | undefined;
+  private maxBytes: number | undefined;
   private readonly sizeCalculator: (value: unknown) => number;
   private totalSize: number = 0;
 
@@ -60,6 +76,7 @@ export class LRUCache<T> {
     this.cache = new Map();
     this.maxSize = options.maxSize ?? 100;
     this.ttl = options.ttl;
+    this.maxBytes = options.maxBytes;
     this.sizeCalculator =
       options.sizeCalculator ??
       (value => {
@@ -69,6 +86,68 @@ export class LRUCache<T> {
           return 1;
         }
       });
+  }
+
+  /**
+   * Reconfigure bounds at runtime. maxSize / ttl / maxBytes are updated in
+   * place and over-capacity entries are evicted. Used by callers that key a
+   * cache by namespace alone (so a policy change reconfigures rather than
+   * strands the old cache — see #124). The size estimator cannot change
+   * after construction (existing entries keep their recorded size); pass a
+   * correct {@link LRUCacheOptions.sizeCalculator} up front.
+   */
+  reconfigure(options: {
+    maxSize?: number;
+    ttl?: number;
+    maxBytes?: number;
+  }): void {
+    if (options.maxSize !== undefined) {
+      this.maxSize = Number.isFinite(options.maxSize)
+        ? Math.max(0, Math.floor(options.maxSize))
+        : this.maxSize;
+    }
+    if ('ttl' in options) {
+      const configuredTtl = options.ttl;
+      this.ttl =
+        configuredTtl !== undefined &&
+        Number.isFinite(configuredTtl) &&
+        configuredTtl > 0
+          ? configuredTtl
+          : undefined;
+    }
+    if ('maxBytes' in options) {
+      const configuredMaxBytes = options.maxBytes;
+      this.maxBytes =
+        configuredMaxBytes !== undefined &&
+        Number.isFinite(configuredMaxBytes) &&
+        configuredMaxBytes > 0
+          ? configuredMaxBytes
+          : undefined;
+    }
+    this.evictToFit();
+  }
+
+  /**
+   * Evict oldest entries until both the count and byte bounds are satisfied.
+   * `protectedKey` is never evicted by this call — used by {@link set} so a
+   * freshly inserted entry that alone exceeds `maxBytes` is retained rather
+   * than immediately dropped (the work to produce it has already happened).
+   */
+  private evictToFit(protectedKey?: string): void {
+    while (
+      this.cache.size > this.maxSize ||
+      (this.maxBytes !== undefined && this.totalSize > this.maxBytes)
+    ) {
+      const firstKey = this.cache.keys().next().value;
+      if (firstKey === undefined || firstKey === protectedKey) {
+        break;
+      }
+      const firstEntry = this.cache.get(firstKey);
+      if (firstEntry) {
+        this.totalSize -= firstEntry.size;
+      }
+      this.cache.delete(firstKey);
+    }
   }
 
   /**
@@ -121,17 +200,10 @@ export class LRUCache<T> {
     this.cache.set(key, entry);
     this.totalSize += size;
 
-    // If capacity is exceeded, evict the oldest entry (the first one in the Map)
-    while (this.cache.size > this.maxSize) {
-      const firstKey = this.cache.keys().next().value;
-      if (firstKey !== undefined) {
-        const firstEntry = this.cache.get(firstKey);
-        if (firstEntry) {
-          this.totalSize -= firstEntry.size;
-        }
-        this.cache.delete(firstKey);
-      }
-    }
+    // Evict oldest entries until both the count and byte bounds hold. The
+    // freshly inserted entry is protected: an item larger than maxBytes is
+    // retained alone rather than evicted to make room for itself.
+    this.evictToFit(key);
   }
 
   /**
@@ -185,12 +257,14 @@ export class LRUCache<T> {
     size: number;
     maxSize: number;
     totalSize: number;
+    maxBytes: number | undefined;
     ttl: number | undefined;
   } {
     return {
       size: this.cache.size,
       maxSize: this.maxSize,
       totalSize: this.totalSize,
+      maxBytes: this.maxBytes,
       ttl: this.ttl,
     };
   }
--- a/packages/core/src/cache.ts
+++ b/packages/core/src/cache.ts
@@ -189,6 +189,13 @@ export class LRUCache<T> {
       this.cache.delete(key);
     }
 
+    // maxSize: 0 is a documented zero-capacity cache. The freshly inserted
+    // key is protected below only so one byte-oversized value can be retained;
+    // it must not bypass the independent entry-count bound.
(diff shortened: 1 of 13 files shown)


packages/core/src/cache.ts
```diff
@@ -71,6 +88,68 @@ export class LRUCache<T> {
...
+   * freshly inserted entry that alone exceeds `maxBytes` is retained rather
+   * than immediately dropped (the work to produce it has already happened).
+   */
+  private evictToFit(protectedKey?: string): void {
+    while (
+      this.cache.size > this.maxSize ||
+      (this.maxBytes !== undefined && this.totalSize > this.maxBytes)
+    ) {
+      const firstKey = this.cache.keys().next().value;
+      if (firstKey === undefined || firstKey === protectedKey) {
```
kookyleo, line 142: The protected key also bypasses the entry-count bound, not only the byte bound. Reproduction: new LRUCache({ maxSize: 0, maxBytes: 100, sizeCalculator: () => 1 }).set("k", "v") leaves size === 1. maxSize: 0 is documented and used as zero capacity, so protect the inserted key only for the single-oversized-byte case (or short-circuit zero capacity), and add a regression test.

packages/renderers/rn/src/DiagramNode.tsx
```diff
@@ -78,7 +78,11 @@ export const DiagramNode: React.FC<DiagramNodeProps> = ({
       ),
     [diagramConfig, globalCache, node.engine]
   );
-  const diagramCache = getRendererCache<CachedDiagramResult>('diagram', cachePolicy);
+  const diagramCache = getRendererCache<CachedDiagramResult>(
+    'diagram',
```
kookyleo: This namespace is shared by every engine, but cachePolicy is resolved per engine. Rendering a dot node with maxSize: 1 therefore reconfigures and evicts entries created under a mermaid maxSize: 100 policy; the last rendered engine wins. Scope the stable namespace at least by normalizedEngine (while still reconfiguring that engine in place), and test two engines with different policies.

packages/renderers/rn/src/Supramark.tsx
```diff
@@ -44,13 +44,106 @@ import {
...
+    for (const key of Object.keys(record)) {
+      deepFreezeAst(record[key]);
+    }
+    for (const sym of Object.getOwnPropertySymbols(record)) {
+      deepFreezeAst(record[sym]);
+    }
+  }
+  // Object.freeze is idempotent and a no-op on non-objects; safe to call.
+  try {
+    Object.freeze(value as object);
```
kookyleo: This is the current required-check failure: Test & Build (20.x) reports @typescript-eslint/no-unnecessary-type-assertion here. value has already been narrowed to object, so Object.freeze(value) should satisfy the rule.

packages/renderers/rn/src/Supramark.tsx
```diff
@@ -44,13 +44,106 @@ import {
...
 type RenderedNode = React.ComponentProps<typeof Text>['children'];
 
+// Dev mode: __DEV__ in React Native bundles, NODE_ENV elsewhere (web/tests).
+// Deep-freeze of the shared cached AST only runs in dev so production keeps
+// the freeze cost off the render path while the read-only contract holds.
+const globalRecord = globalThis as Record<string, unknown> & {
+  process?: { env?: Record<string, string | undefined> };
+};
+const isDevMode: boolean =
+  (typeof globalRecord.__DEV__ !== 'undefined' && globalRecord.__DEV__ === true) ||
```
kookyleo: When __DEV__ is explicitly false, this OR expression can still fall through to process.env.NODE_ENV and become true if that value is absent or non-production. Since the comment says RN __DEV__ is authoritative, prefer a conditional: use __DEV__ when it is defined, and consult NODE_ENV only when it is not.

packages/renderers/rn/src/renderCache.ts
```diff
@@ -5,36 +5,60 @@ export interface RendererCachePolicy {
...
+   */
+  maxBytes?: number;
 }
 
 /** Cache policy input accepted by existing Supramark diagram configuration. */
 export interface RendererCachePolicyInput {
   enabled?: boolean;
   maxSize?: number;
   ttl?: number;
+  maxBytes?: number;
```
kookyleo, line 21: maxBytes is accepted internally, but the public SupramarkDiagramEngineConfig.cache and SupramarkDiagramConfig.defaultCache types in packages/core/src/ast.ts still expose only enabled/maxSize/ttl. A TypeScript host therefore cannot set the documented byte bound without a cast. Please add maxBytes to both public cache shapes (or remove the claim that hosts can tune it).

packages/core/src/feature.ts
```diff
@@ -1653,10 +1653,14 @@ export interface SupramarkConfig {
...
- * host runtime (@supramark/rn / @supramark/web) enables a process-level runtime cache
- * — reused by parsed documents and normalized diagram SVGs across scenarios like
- * virtual-list remounts. This cache is bounded by default (by entry count) and can be
- * adjusted or disabled via `diagram.defaultCache` / `diagram.engines[engine].cache`.
+ * RN host runtime (@supramark/rn) enables a process-level runtime cache — reused by
+ * parsed documents and normalized diagram SVGs across scenarios like virtual-list
+ * remounts. This cache is bounded by default (both entry count and a total-byte cap)
+ * and can be adjusted or disabled via `diagram.defaultCache` / `diagram.engines[engine].cache`.
+ *
+ * Note: the web runtime (@supramark/web) does NOT maintain a document or diagram
```
kookyleo: This becomes stale once #170 lands: that PR adds a real process-level Web diagram-result cache. The accurate split is RN = parsed-document + diagram cache; Web = diagram cache only. Please reconcile this wording before both changes merge.

kookyleo (changes requested):
The implementation addresses the original #124 measurements, and the 64 focused cache/RN tests pass locally. The current head is still blocked by the required Node 20 lint:types failure, and the inline comments identify additional cache-bound and policy-isolation regressions that need correction. One unanchored blocker: resolveDocumentCachePolicy() returns only maxSize and ttl when caching is enabled solely by per-engine policies; it drops maxBytes, so the parsed-document cache remains byte-unbounded in exactly that supported configuration. Please aggregate the strictest maxBytes there and cover it with a test. Also update the cache documentation for the now-approved #170: Web has a diagram-result cache but no parsed-document cache.

kookyleo (approved):
All requested cache-bound, policy-isolation, public typing, dev-mode, lint, and documentation issues are addressed at d3b1885, including the unanchored strictest maxBytes aggregation case. I verified 68 focused tests, the full core + RN suites (206 tests / 442 expectations), both package builds, targeted ESLint, and lint:types locally. All current required and auxiliary GitHub checks are green.

PR #166, opened 2026-08-04
fix(core): treat missing renderers as production-mode validation failure

### LINKED ISSUE #164: [core] validateFeature passes features with no renderers field in every mode (incl. production)
## Summary

`validateFeature` has no renderer check in basic/strict mode, and the only renderer check in production mode is gated by `if ('renderers' in feature)` — so a `Partial<SupramarkFeature>` that omits the `renderers` key entirely passes validation in ALL modes with zero renderers. The immediately-following `testing` check uses `if (!('testing' in feature) || !feature.testing)` (correctly treating absence as a problem), making the `renderers` gating inconsistent with the `testing` gating.

## Location

`packages/core/src/feature.ts`, production block:

```ts
// In production mode, at least one renderer should be defined
if ('renderers' in feature) {          // ← absence skips the check entirely
  const renderers = feature.renderers as RendererDefinitions<SupramarkNode>;
  const hasRenderer = renderers && (renderers.rn || renderers.web || renderers.cli);
  if (!hasRenderer) {
    errors.push({ ... 'renderers-required-production' ... });
  }
}

// In production mode, tests are recommended
if (!('testing' in feature) || !feature.testing) {   // ← absence treated as a problem
  errors.push({ ... 'testing-recommended-production' ... });
}
```

There is no `renderers` check in the basic/strict section above, so absence passes every mode.

## Scenario

A feature object `{ metadata, syntax }` (no `renderers`, no `testing`) submitted to `validateFeature(feature, { production: true })`:
- `testing` check fires (warning) — correct.
- `renderers` check is skipped (no `renderers` key) — should fire `renderers-required-production` but does not.

## Why CI misses it

The validator IS the gate. All real features in-repo include `renderers`, so the gap is latent — a malformed or third-party feature slips through.

## Fix sketch

Replace `if ('renderers' in feature)` with `if (!('renderers' in feature) || !feature.renderers || !(feature.renderers.rn || feature.renderers.web || feature.renderers.cli))`, treating absence as the defect — mirroring the `testing` check.

## Severity

Med (latent) — the validation gate has a hole; in-repo features all carry renderers so no active breakage, but the inconsistency with `testing` confirms it is unintentional.

mybee:
## Summary

- `validateFeature` gated its only renderer check on `if ('renderers' in feature)`, so a `Partial<SupramarkFeature>` omitting the `renderers` key passed production validation with zero renderers.
- The adjacent `testing` check already treats absence as a defect (`!('testing' in feature) || !feature.testing`). This PR makes `renderers` consistent: absence now fires `renderers-required-production`.
- Adds a regression test covering a feature with no `renderers` key submitted in production mode.

Closes #164.

## Test plan

- [x] `bun test packages/core/__tests__/validateFeature.test.ts` — 16 pass
- [x] `tsc --noEmit` (core) — clean
- [ ] CI: `bun run quality`

### DIFF
--- a/packages/core/__tests__/validateFeature.test.ts
+++ b/packages/core/__tests__/validateFeature.test.ts
@@ -186,6 +186,27 @@ describe('validateFeature', () => {
       const result = validateFeature(feature as unknown as Parameters<typeof validateFeature>[0]);
       expect(result.errors.some(e => e.code === 'ast-interface-fields-defined')).toBe(true);
     });
+
+    it('warns when renderers is missing or has no platform renderer', () => {
+      const feature = {
+        metadata: {
+          id: '@supramark/feature-test',
+          name: 'Test Feature',
+          version: '1.0.0',
+        },
+        syntax: {
+          ast: {
+            type: 'test_node',
+          },
+        },
+        // renderers key omitted entirely
+      };
+
+      const result = validateFeature(feature as unknown as Parameters<typeof validateFeature>[0]);
+      expect(result.errors.some(e => e.code === 'renderers-required')).toBe(true);
+      // basic mode treats it as a warning, not a failure
+      expect(result.valid).toBe(true);
+    });
   });
 
   describe('info-level checks', () => {
@@ -264,6 +285,7 @@ describe('validateFeature', () => {
             type: 'test_node',
           },
         },
+        renderers: { rn: { Component: () => null } },
       };
 
       const result = validateFeature(feature as unknown as Parameters<typeof validateFeature>[0], { strict: true });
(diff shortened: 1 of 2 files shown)


kookyleo (approved):
Re-reviewed head `c21ee233`. The follow-up commits resolve both concerns from my earlier review:

- missing/empty renderers now produce `renderers-required` in basic/strict mode
- production mode emits only `renderers-required-production`, without a duplicate warning
- the misleading testing-check comparison was removed

- `bun test packages/core/__tests__/validateFeature.test.ts`: 17 passed
- core declaration build: passed
- ESLint on both changed files: passed

All 22 GitHub checks are green. No blocking findings.

PR #167, opened 2026-08-04
fix(engines): consume options.timeout so hung renders don't block the document

### LINKED ISSUE #162: [engines] timeoutMs is never consumed — a hung wasm render blocks the whole document forever
## Summary

`buildDiagramRenderOptions` (both web and RN) forwards `engineConfig.timeoutMs` into `options.timeout`, but no engine reads `options.timeout` and there is no `Promise.race` / timeout wrapper anywhere in the render path. A hung wasm render (infinite loop on a malformed graph, elk bridge deadlock) makes `preRenderAll`'s `Promise.all` never settle, so the document containing it is stuck at `pending: true` forever — including every other healthy diagram in the same document.

## Location

- Forwarding: `packages/renderers/web/src/Supramark.tsx:1830` and `packages/renderers/rn/src/DiagramNode.tsx:300` — `base.timeout = engineConfig.timeoutMs`
- No consumer: `grep -n timeout` across `packages/engines/src/{engine,mermaid,echarts,vega-lite,mathjax,graphviz,web}.ts` returns zero reads of `options.timeout` / `timeoutMs`.
- No race: `packages/renderers/web/src/Supramark.tsx:1718` `preRenderAll` is a bare `Promise.all(taskList.map(t => engine.render(...)))` with no per-task timeout.

## Scenario

- Host sets `config.diagram.engines.d2.timeoutMs = 8000`.
- A malformed d2 graph hangs inside `loadWebD2Render`'s `convert()`.
- `preRenderAll` awaits `Promise.all`, which never settles → `setParsedDocument` never runs → `onRenderStateChange` stuck at `pending: true` for the **whole** document (not just the hung diagram — every other healthy diagram in that doc is dragged down by the `Promise.all`). The effect's `cancelled` flag cannot help, because the promise never settles.

## Why CI misses it

The option is typed and forwarded correctly end-to-end; the bug is that nobody consumes it. No test asserts a timeout fires.

## Fix sketch

Either consume `options.timeout` in `engine.render` via `Promise.race` against a rejection timer, or wrap each `engine.render(...)` call in `preRenderAll` with a per-task timeout that resolves to a `render_error` result on expiry.

## Severity

Med — requires a hang to trigger (rare), but unrecoverable when it happens and affects the whole document, not just the hung diagram.

mybee:
## Summary

`engineConfig.timeoutMs` was forwarded as `options.timeout` by both web and RN renderers, but no engine consumed it and there was no `Promise.race` anywhere in the render path. A hung wasm render (d2 `convert()` on a malformed graph, elk bridge deadlock) left `preRenderAll`'s `Promise.all` unsettled forever, pinning the **whole** document at `pending: true` — including every other healthy diagram in it.

This consumes `options.timeout` (plus the `timeoutMs` alias) centrally in `LocalDiagramEngine.render` via `Promise.race` against a `RenderTimeoutError`. On expiry the engine returns a `render_error` result so the document settles. The underlying hung promise is left pending since wasm renders can't be cancelled; its eventual resolution is dropped.

Centralizing in `engine.render` covers both consumers:
- web `preRenderAll` (`Promise.all` of `engine.render`)
- RN per-node rendering in `DiagramNode`

## Why centralize in the engine, not in `preRenderAll`

`preRenderAll` is web-only. RN renders diagrams one node at a time in `DiagramNode.tsx`, so wrapping only `preRenderAll` would leave RN unprotected. The engine is the single choke point both consumers share.

## Test plan

- [x] New `packages/engines/__tests__/engine-render-timeout.test.ts` — 6 cases: hang → `render_error` with `timed out` payload, `timeoutMs` alias, completes-before-timeout returns svg, no-timeout path unaffected, non-positive timeouts ignored, thrown render errors still surface normally.
- [x] `bun --filter @supramark/engines test` — 26 pass.
- [x] `tsc -p packages/engines/tsconfig.json` — clean.
- [x] `eslint` on changed files — clean.

Closes #162


kookyleo (approved):
Verified the timeout implementation after rebasing onto current main and reconciling it with #168 loader retries. Local validation: engines/Web/RN suites (181 tests, 396 expectations), all three TypeScript builds, repository type-boundary lint, targeted ESLint, and Prettier. The exact head d5bd5e29 is mergeable, has no unresolved threads, and all 18 current GitHub checks are green.

PR #168, opened 2026-08-04
fix(engines): clear cached loader promise on rejection

### LINKED ISSUE #161: [engines] rejected loader promise cached forever — single transient load failure permanently bricks the engine
## Summary

`getEchartsRender` / `getVegaLiteRender` / `getPlantumlRender` / `getD2Render` / `getGraphvizAdapter` store the promise from `loadRender()` / `loadAdapter()` in a private field and return the cached field on every subsequent call. If that promise rejects once (chunk-load failure, CDN timeout, transient wasm init error), the **rejected** promise is cached and re-awaited forever — every later `engine.render({engine:'echarts', ...})` returns a `render_error` result. The engine is permanently disabled for the rest of the app's process lifetime; remounting the component does nothing, only a full page reload recovers.

## Location

- `packages/engines/src/engine.ts:155-196` — the five `getXxxRender()` getters:
  ```ts
  if (!this.echartsRenderPromise) {
    this.echartsRenderPromise = this.options.echarts.loadRender();
  }
  return this.echartsRenderPromise;
  ```
  After a reject the field is still a (rejected) Promise object — truthy — so `!this.echartsRenderPromise` stays false and it is never re-assigned.
- `packages/engines/src/web.ts:97` — `createWebDiagramEngine()` returns `new LocalDiagramEngine({ echarts: { loadRender: loadEchartsSvgRender }, ... })` (the instance holding those fields).
- `packages/renderers/web/src/Supramark.tsx:112` + `packages/renderers/rn/src/DiagramNode.tsx:35` — `const defaultDiagramEngine = createXxxDiagramEngine()` at module top level (singleton for the app lifetime).

## Repro

1. Web Supramark rendering an echarts (or vega-lite / plantuml / d2 / graphviz) block.
2. Simulate a transient `loadEchartsSvgRender()` rejection once (throttle network → chunk 404 once).
3. Restore the network, re-render, remount the component, navigate away and back.
4. The engine never renders again until a full page reload; every render returns `render_error`.

## Why CI misses it

tsc sees a valid `Promise<DiagramRenderFn>` field. eslint has no rule for this. Tests inject a pre-resolved `render` fn, so no test exercises a rejected `loadRender`. The `render` try/catch converts the rejection into a stream of `render_error` results that look like bad input rather than a stuck loader.

## Fix sketch

In each getter, clear the cached field when the stored promise rejects, so the next call retries the load:

```ts
if (!this.echartsRenderPromise) {
  const p = this.options.echarts.loadRender();
  p.catch(() => { if (this.echartsRenderPromise === p) this.echartsRenderPromise = null; });
  this.echartsRenderPromise = p;
}
return this.echartsRenderPromise;
```

## Severity

High — a single transient load failure permanently disables an engine for the app lifetime, with no recovery short of full reload.

mybee:
## Summary

- `getEchartsRender` / `getVegaLiteRender` / `getPlantumlRender` / `getD2Render` / `getGraphvizAdapter` cached the `loadRender()` / `loadAdapter()` promise in an instance field and returned it on every subsequent call. A single rejection (chunk-load failure, CDN timeout, transient wasm init error) was re-awaited forever, permanently disabling the engine for the app lifetime — the default engine is a module-level singleton, so remounting the component does nothing; only a full page reload recovers.
- Drop the cached promise on rejection so the next render retries the load. The identity guard avoids clearing a newer promise if a retry already replaced the field.

## Test plan

- [x] `bun test packages/engines/__tests__/` — 23 pass, including 3 new in `engine-loader-retry.test.ts`:
  - `loadRender` rejects once → next render retries the load and succeeds
  - successful `loadRender` is cached across renders (no re-load on hit)
  - graphviz `loadAdapter` rejects once → next render retries and succeeds
- [x] `tsc -p packages/engines/tsconfig.json --noEmit` — clean
- [x] `eslint packages/engines/src/engine.ts packages/engines/__tests__/engine-loader-retry.test.ts --max-warnings 0` — clean

Closes #161.

### DIFF
--- a/packages/engines/__tests__/engine-loader-retry-graphviz.test.ts
+++ b/packages/engines/__tests__/engine-loader-retry-graphviz.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, it, mock } from 'bun:test';
+
+// The injected-`loadAdapter` test in `engine-loader-retry.test.ts` bypasses
+// `createWebGraphvizAdapterLoader()`, which is precisely the production path
+// that defeated the #161 fix: the default loader caches its own
+// `adapterPromise` with no clear-on-rejection, so once `Graphviz.load()`
+// rejects the engine stays bricked for life even after `engine.ts`'s outer
+// cache is cleared. This file exercises the DEFAULT wiring
+// (`createWebDiagramEngine({})` → `createWebGraphvizAdapterLoader` →
+// `loadWebGraphvizAdapter` → `@actrium/graphviz-anywhere-web`'s
+// `Graphviz.load()`) with a module mock that counts real load attempts.
+
+let loadCalls = 0;
+
+mock.module('@actrium/graphviz-anywhere-web', () => ({
+  __esModule: true,
+  Graphviz: {
+    load: () => {
+      loadCalls += 1;
+      if (loadCalls === 1) {
+        return Promise.reject(new Error('transient wasm load failure'));
+      }
+      return Promise.resolve({
+        layout: () => '<svg>gv</svg>',
+        version: () => 'mock-1.0',
+      });
+    },
+  },
+}));
+
+const { createWebDiagramEngine } = await import('../src/web');
+
+describe('createWebGraphvizAdapterLoader retry on rejection (#161, default wiring)', () => {
+  it('retries Graphviz.load on the next render after a transient failure', async () => {
+    // No injected `loadAdapter` — exercise the production default wiring.
+    const engine = createWebDiagramEngine();
+    const code = 'digraph { a -> b }';
+
+    const first = await engine.render({ engine: 'dot', code });
+    expect(first.success).toBe(false);
+    expect(loadCalls).toBe(1);
+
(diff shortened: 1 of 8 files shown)


kookyleo (approved):
Re-reviewed head `32ae8ac6`. The follow-up commits address the previously blocking cache layers:

- Graphviz now clears rejected promises in the default web/RN loader closures and both published adapter entrypoints
- PlantUML clears both deferred wasm and Graphviz-bridge promises
- MathJax follows the same retry-on-rejection contract
- the default-wiring Graphviz regression test reaches the real inner cache
- per-loader tests cover ECharts, Vega-Lite, PlantUML, D2, and Graphviz
- the identity-guard comment now describes the current invariant accurately

- `bun test packages/engines/__tests__/`: 27 passed / 0 failed
- engines TypeScript build: passed
- ESLint on all 8 changed files: passed

All GitHub checks are green. A real browser may still cache a failed dynamic import at the module-map/bundler layer, but this PR correctly restores retryability for every cache controlled by Supramark and does not overpromise beyond that. No blocking findings.

PR #169, opened 2026-08-04
fix(core): make FeatureRegistry.register idempotent for HMR (#165)

### LINKED ISSUE #165: [features] module-level FeatureRegistry.register() throws on HMR re-import
## Summary

Every diagram / container / card / code-highlight feature calls `FeatureRegistry.register(featureX)` at module top level. `register` throws `Feature "X" is already registered` on duplicate (raw `throw`, no dedup guard). In a Vite dev server with HMR, editing a feature file re-executes the top-level call on re-import, so the throw fires, breaks the HMR update with an uncaught error (error overlay), and the hot-reload cycle does not recover until a full page reload.

This also violates the project's own `CLAUDE.md` rule: "不要添加全局副作用（如模块级注册）".

## Location

- Top-level register in 10 features:
  - `packages/features/diagrams/mermaid/src/feature.ts:72`
  - `packages/features/diagrams/d2/src/feature.ts:89`
  - `packages/features/diagrams/dot/src/feature.ts:71`
  - `packages/features/diagrams/echarts/src/feature.ts:91`
  - `packages/features/diagrams/plantuml/src/feature.ts:102`
  - `packages/features/diagrams/vega-lite/src/feature.ts:107`
  - `packages/features/containers/html-page/src/feature.ts:225`
  - `packages/features/containers/map/src/feature.ts:237`
  - `packages/features/cards/vison/src/feature.ts:249`
  - `packages/features/main/code-highlight/src/feature.ts:102`
- `packages/core/src/feature.ts:1248` — `if (this.features.has(id)) throw new Error(...)`, no dedup.

## Scenario

1. `bun run feature:preview:web` (or any Vite app importing a feature module).
2. Edit the feature file (e.g. tweak `examples`).
3. Vite HMR re-imports the module → top-level `register(...)` runs again → `this.features.has(id)` is true → throws.
4. Uncaught throw during module eval → Vite error overlay, HMR update fails. Editing again / fixing does not recover until a full reload, because the registry still holds the first registration and the new module instance throws on init.

## Why CI misses it

tsc / eslint don't flag module-level side effects. Tests call `FeatureRegistry.clear()` in setup, so the duplicate-registration path never fires in the test run.

## Fix sketch

Either make `register` idempotent (overwrite-or-skip when already registered, or dedup by feature-object identity), or move registration out of module top level — have the host explicitly register the features it imports (matches the passive-rendering model in `CLAUDE.md`). The latter aligns with the "no global side effects" rule.

## Severity

Med — dev-only (HMR), no production impact, but it breaks the edit-loop for anyone working on a feature package.

mybee:
## Summary

- Fixes #165. `FeatureRegistry.register()` threw `Feature "X" is already registered` on any duplicate id, and every diagram/container/card/code-highlight feature calls `register()` at module top level. Under Vite HMR a hot-update re-evaluates that top-level call with a fresh feature object, so the throw fired during module evaluation, broke the hot-reload loop with an uncaught error, and only a full page reload recovered it.
- `register()` is now idempotent: re-registering the same object reference is a no-op, and a fresh object with an existing id overwrites the stale entry instead of throwing. The dev loop recovers and the edited feature is reflected, while the no-throw contract the passive-rendering model expects is preserved.
- Adds a regression test covering both the same-ref no-op and the HMR fresh-object overwrite paths.

## Why the minimal fix

The issue lists two approaches — idempotent `register()` vs. moving registration out of module top level. This PR takes the idempotent route: it is a single-site change in core that all 13 module-level `register()` call sites benefit from automatically, with no breaking change to host import ergonomics. The host-driven explicit feature injection model from `CLAUDE.md` is unchanged.

## Test plan

- [x] `bun run test:core` — 90 pass
- [x] `tsc --noEmit` in `packages/core` — clean
- [x] New regression test asserts no throw on duplicate id and that the fresh object replaces the stale one

### DIFF
--- a/packages/core/__tests__/feature-config.test.ts
+++ b/packages/core/__tests__/feature-config.test.ts
@@ -114,6 +114,17 @@ describe('Feature configuration system', () => {
 
       expect(FeatureRegistry.get('@test/feature-a')).toBe(refreshed);
       expect(FeatureRegistry.list()).toHaveLength(1);
+
+      // Subsequent HMR-style re-registrations of the same id stay quiet: the
+      // duplicate-id warning fires once per session, not once per save.
+      const refreshedAgain = createTestFeature('@test/feature-a', 'test-a-updated-2');
+      const spy3 = jest.spyOn(console, 'warn').mockImplementation(() => {});
+      expect(() => FeatureRegistry.register(refreshedAgain)).not.toThrow();
+      expect(spy3).not.toHaveBeenCalled();
+      spy3.mockRestore();
+
+      expect(FeatureRegistry.get('@test/feature-a')).toBe(refreshedAgain);
+      expect(FeatureRegistry.list()).toHaveLength(1);
     });
   });
 
--- a/packages/core/__tests__/feature-config.test.ts
+++ b/packages/core/__tests__/feature-config.test.ts
@@ -80,6 +80,22 @@ describe('Feature configuration system', () => {
       });
     });
 
+    it('re-registering an existing id is idempotent and does not throw (HMR re-import)', () => {
+      const feature = createTestFeature('@test/feature-a', 'test-a');
+      FeatureRegistry.register(feature);
+
+      // Re-registering the same object reference is a no-op.
+      expect(() => FeatureRegistry.register(feature)).not.toThrow();
+
+      // A fresh object with the same id (what Vite HMR produces on hot
+      // update) replaces the previous entry instead of throwing.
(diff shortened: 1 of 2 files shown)


kookyleo (approved):
Verified the HMR duplicate-registration path. Same-reference registration is a silent no-op, a fresh object with the same id replaces the stale entry without throwing, and duplicate warnings are deduplicated per id and reset by clear(). The focused 22-test suite, TypeScript check, targeted ESLint, and the full CI matrix all pass.

PR #170, opened 2026-08-04
fix(web): honor engineConfig.cache via a real diagram cache (#163)

### LINKED ISSUE #163: [renderers/web] no runtime cache; engineConfig.cache is a silent no-op on web
## Summary

The web renderer has no module-level runtime cache — `grep` for `AsyncRendererCache | getRendererCache | renderCache | LRUCache` across `packages/renderers/web/src/` returns zero hits. `buildDiagramRenderOptions` forwards `engineConfig.cache` into render `options`, but no web engine reads `options.cache` either. So a host setting `diagram.engines.mermaid.cache = { enabled: true, maxSize: 50 }` on web gets nothing — the option is silently dropped — while the RN renderer honors the same field to build a real LRU (`packages/renderers/rn/src/renderCache.ts`, `DiagramNode.tsx:81`).

Consequence: virtual-list recycling, tab switches, and React strict-mode remounts re-render every diagram from scratch on web — no SVG reuse. A doc with 20 mermaid blocks re-runs all wasm work on every scroll-back.

## Location

- No cache: `packages/renderers/web/src/**` (all 10 files) — zero matches for the cache primitives above.
- Forwarding without consumption: `packages/renderers/web/src/Supramark.tsx` `buildDiagramRenderOptions` (~line 1830).
- RN parity for reference: `packages/renderers/rn/src/renderCache.ts` + `DiagramNode.tsx:81` (diagram SVG cache) + `Supramark.tsx` (parsed-document cache).

## Why CI misses it

The asymmetry is architectural (web has no `renderCache.ts` at all), so no test asserts web caching behavior. `buildDiagramRenderOptions` forwarding the field makes it look wired.

## Fix sketch

Either (a) port `AsyncRendererCache` to web and honor `engineConfig.cache` the way RN does, or (b) if web is intentionally cache-less, document `engineConfig.cache` as an RN-only facility and have `buildDiagramRenderOptions` stop forwarding (or warn) on web. Either resolves the silent no-op.

## Severity

Med — silent drop of a documented option + a perf cliff on remount-heavy hosts.

mybee:
## Summary

The web renderer had no module-level runtime cache — `AsyncRendererCache | getRendererCache | renderCache | LRUCache` returned zero hits across `packages/renderers/web/src/`. `buildDiagramRenderOptions` forwarded `engineConfig.cache` into render `options`, but no web engine read `options.cache` either, so a host setting `diagram.engines.mermaid.cache = { enabled: true, maxSize: 50 }` on web got nothing — the option was silently dropped (#163). On remount-heavy hosts (virtual-list recycling, tab switches, React strict-mode remounts) every diagram re-ran the full wasm render from scratch — a doc with 20 mermaid blocks re-ran all wasm work on every scroll-back.

### Fix

Port the RN renderer's `AsyncRendererCache` to web ([packages/renderers/web/src/renderCache.ts](packages/renderers/web/src/renderCache.ts)) and consume the resolved cache policy inside `preRenderAll`, mirroring `DiagramNode.tsx` on RN:

- `resolveDiagramCachePolicy(engine > diagram.defaultCache > global options.cache)`
- `getRendererCache('diagram', policy)` — shared, policy-keyed LRU
- `getOrCreate` keyed by engine identity + engine + code + resolved options, so a cached result cannot cross engine instances or option variants (e.g. a different PlantUML server URL)
- retain only successful results (`result.success`) so a transient engine error is retried on the next mount instead of being served from cache

`buildRenderKey` now uses the shared `stableSerialize`, which also handles cycles and non-plain option values (Date / Map / Set) that previously collided on the same key.

### Scope note

The document-level (parsed-document) cache is intentionally **not** ported. Web's `ParsedDocument` bundles the `rendered` map, so caching it safely would require fingerprinting the whole diagram config in the cache key; the diagram cache alone resolves both the silent no-op (the headline bug) and the remount perf cliff (the headline symptom). Parse cost on remount is negligible relative to wasm rendering.

## Test plan

- [x] `bun --filter @supramark/web build` (tsc --emitDeclarationOnly) passes
- [x] `bun run build` (repo-wide) passes
- [x] eslint clean on changed files
- [x] `bun test` in `packages/renderers/web` — 54 pass (added 24 + 5 cache tests, no regressions)

New tests (`packages/renderers/web/__tests__/`):
- `renderCache.test.ts` — `stableSerialize` collision/cycle coverage, policy resolution precedence, `getRendererCache` sharing, `AsyncRendererCache` in-flight dedup + error-retry contract
- `SupramarkCache.test.tsx` — remount reuses the engine render when cache is enabled; re-renders when disabled; per-engine cache honored without global cache; swapping the engine instance bypasses the prior cache

Closes #163.

### DIFF
--- a/packages/renderers/web/__tests__/SupramarkCache.test.tsx
+++ b/packages/renderers/web/__tests__/SupramarkCache.test.tsx
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import { afterEach, beforeEach, describe, expect, setSystemTime, test } from 'bun:test';
 import React from 'react';
 import { Window } from 'happy-dom';
 import { createRoot, type Root } from 'react-dom/client';
@@ -40,6 +40,8 @@ afterEach(async () => {
     root = null;
   }
   browser.document.body.replaceChildren();
+  // Reset any mocked system time so it can't leak into the next test.
+  setSystemTime();
 });
 
 function createContainer(): TestContainer {
@@ -103,17 +105,35 @@ async function renderDocument(config: unknown): Promise<void> {
         })
       )
     );
-    // Let the async effect (expandOpaqueContainers + preRenderAll) settle.
-    await Promise.resolve();
-    await Promise.resolve();
-    await Promise.resolve();
   });
+  // Let the async effect (expandOpaqueContainers + preRenderAll) settle.
+  await settle();
 }
 
 async function unmount(): Promise<void> {
   if (root) {
     await act(async () => root?.unmount());
     root = null;
+  }
+}
+
+// Flush microtasks until engineState.renderCalls has been stable for a few
+// consecutive ticks. Covers both the cache-miss path (count increments once
+// then settles) and the cache-hit path (count never moves) without depending
+// on a fixed number of awaits — robust to the effect gaining an extra async
+// tick. Uses a tick counter rather than Date.now() so mocked time can't hang
+// the loop.
+async function settle(stableTicks = 3, maxTicks = 1000): Promise<void> {
+  let last = engineState.renderCalls;
+  let stable = 0;
+  for (let i = 0; i < maxTicks; i++) {
+    await Promise.resolve();
+    if (engineState.renderCalls === last) {
+      if (++stable >= stableTicks) return;
+    } else {
+      last = engineState.renderCalls;
+      stable = 0;
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


kookyleo (approved):
Verified the cache behavior end to end. The Web package passes all 56 tests, its declaration build and targeted ESLint checks pass, and the full CI matrix is green. Policy precedence, in-flight deduplication, successful-result retention, engine-instance isolation, and TTL expiry are all covered and behave as intended.

PR #174, opened 2026-08-08
fix(graphviz-rn): sync generated native resolver output

boomsi:
## Summary

- regenerate the tracked CommonJS, ESM, and TypeScript outputs from the current Graphviz RN source
- bring Git and `link:` consumers onto the TurboModule-first, NativeModules fallback, and lazy linking-error behavior merged in #17
- export the generated `resolveNative` declaration consistently with the source API

This is a distribution-output sync only; the source implementation and its tests were already merged in #17.

## Test plan

- [x] ran `bun run --filter @actrium/graphviz-anywhere-rn build` twice; both generated the identical diff SHA-256
- [x] `bun test crates/graphviz-anywhere/packages/react-native/__tests__` — 7 passed
- [x] `bun run --filter @actrium/graphviz-anywhere-rn typescript`
- [x] targeted ESLint for the Graphviz RN source and tests
- [x] loaded the generated CommonJS entry and verified bridge delegation plus the `resolveNative` export


kookyleo (approved):
Reviewed head `1989b64d`.

This is a clean generated-output sync:

- the tracked CommonJS, ESM, and declaration outputs match the current `src/index.ts`
- rerunning Bob regenerated all three targets with zero git diff
- `resolveNative` is exported consistently and the bridge fallback remains lazy
- both generated JavaScript entry formats delegate default `dot/svg` arguments and `getVersion` correctly

- Graphviz RN source suite: 7 passed / 0 failed
- TypeScript check: passed
- Bob build: passed, reproducible with zero diff
- direct generated-entry smoke: CommonJS + ESM, 2 passed / 8 assertions
- all 15 GitHub checks are green

No blocking findings.

PR #187, opened 2026-08-10
fix(rn-selection): enforce viewport gesture boundaries

### LINKED ISSUE #182: rn-selection: cross-viewport reverse selection (outside-in not clamped)
From PR #159 review at `31697fa8` (both iOS + Android):

Dragging a handle from *outside* the FlatList still selects text *inside* it. Inside→out is now clamped (`resolvePointToSelectionInViewport`), but outside→in is not — the reverse direction still crosses the viewport boundary.

### LINKED ISSUE #183: rn-selection: momentum scroll pauses interaction outside the FlatList
From PR #159 review at `31697fa8` (both iOS + Android):

During FlatList momentum scroll, elements outside the FlatList are unresponsive. Suspected: `SelectionViewport` pauses interaction too broadly during momentum (should be scoped narrower, or only pause the FlatList itself, not the whole root).

kookyleo:
## Summary

- treat root content outside nested selection viewports as an explicit gesture ownership scope
- prevent outside-in handle drags from selecting FlatList content, matching the existing inside-out clamp
- release stale viewport touch ownership when native momentum begins so UI outside the FlatList remains interactive during a fling
- add pure hit-test and component-level responder regressions for both directions

Closes #182
Closes #183

## Verification

- `bun test packages/renderers/rn-selection/src/__tests__` — 291 pass, 0 fail, 711 expectations
- `bun run build` in `packages/renderers/rn-selection` — pass
- `bun run lint` — pass (ESLint, 21 feature packages, CJK)
- `bun run lint:types` — 0 findings
- `bun run quality` — pass
- Prettier and `git diff --check` — pass

The repository root test command also ran this package successfully (291/291). Its aggregate process later failed while core and RN concurrently rebuilt the same markdown WASM and inherited the developer machine's native `-fuse-ld=mold` flag into `rust-lld`; that repository-level command/environment issue is unrelated to these TypeScript-only changes and is being handled separately in #186.

### DIFF
--- a/packages/renderers/rn-selection/src/__tests__/coordinator/SelectionRoot.test.tsx
+++ b/packages/renderers/rn-selection/src/__tests__/coordinator/SelectionRoot.test.tsx
@@ -863,6 +863,93 @@ describe('SelectionRoot responder negotiation', () => {
     expect(activity).toEqual([true, false]);
   });
 
+  test('a handle that starts outside a viewport cannot select into it', () => {
+    const scopedUnits: SelectionUnit[] = [
+      {
+        kind: 'text',
+        unitId: 'outside#0',
+        nodeId: 'outside',
+        text: 'outer',
+        node: NODE,
+      },
+      {
+        kind: 'text',
+        unitId: 'inside#0',
+        nodeId: 'inside',
+        text: 'inner',
+        node: NODE,
+      },
+    ];
+    let store: SelectionStore | null = null;
+
+    const RegisterBlocks: React.FC = () => {
+      const ctx = useSelectionContext();
+      useEffect(() => {
+        store = ctx.store;
+        const disposeOutside = ctx.registerBlock({
+          nodeId: 'outside',
+          unitIds: ['outside#0'],
+          kind: 'text',
+          rect: { x: 10, y: 10, w: 60, h: 20 },
+        });
+        const disposeInside = ctx.registerBlock({
+          nodeId: 'inside',
+          unitIds: ['inside#0'],
+          kind: 'text',
+          rect: { x: 10, y: 100, w: 60, h: 20 },
+          viewportId: 'list',
+        });
+        return () => {
+          disposeOutside();
+          disposeInside();
+        };
+      }, [ctx]);
+      return null;
+    };
+
+    act(() => {
+      renderer = create(
+        <SelectionRoot units={scopedUnits} overlay={false} toolbar={false}>
+          <RegisterBlocks />
+        </SelectionRoot>
+      );
+    });
+    if (store === null) throw new Error('test did not capture the selection store');
+    act(() => {
+      store.beginAt({ nodeId: 'outside', unitId: 'outside#0', offset: 0 });
+      store.extendTo({ nodeId: 'outside', unitId: 'outside#0', offset: 5 });
+      store.commit();
+    });
+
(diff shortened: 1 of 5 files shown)


mybee (approved):
Reviewed the full diff against #182 / #183 and the surrounding source. Both fixes are correct, minimal, and backward-compatible; CI is green. Approving.

## Fix 1 — three-state `gestureViewportIdRef` (outside-in clamp)

`undefined | null | string` is a clean discriminator:
- `undefined` — no block owned the gesture start → `resolvePointToSelection` (all blocks, pre-existing behavior preserved)
- `null` — explicit root-content scope → `resolvePointToSelectionOutsideViewports` (only `viewportId === undefined` blocks)
- `string` — a particular nested viewport → `resolvePointToSelectionInViewport`

`block.viewportId ?? null` correctly maps root blocks (viewportId `undefined`) to the `null` scope in both `scopeGestureToHandle` and the `onTouchStart` long-press path, so the clamp covers handle drags *and* long-press — broader than just #182, which is fine.

Verified no regression and no scope leakage:
- Long-press landing on whitespace still yields `undefined` (containingBlock returns null) → full resolve. Pre-existing behavior, unchanged, out of #182 scope.
- The ref is reset to `undefined` in all three exit paths (`onResponderRelease`, `onHandleRelease`, `cancelGesture`), so no stale scope bleeds into the next gesture.
- `resolvePointToSelectionOutsideViewports` filters on `viewportId === undefined`, which matches the `string | undefined` type on `RegisteredBlock` — no root block is mis-excluded.

Symmetric with the existing inside-out clamp (`resolvePointToSelectionInViewport` filters + clips to `clipRect`).

## Fix 2 — `onMomentumScrollBegin` force-release

Treating momentum-start as the authoritative release for the swallowed Android touch-end is the right call: momentum only begins after the finger lifts, so `touchingRef/draggingRef` are semantically stale by then. Setting both false before `releaseIfIdle()` guarantees the release instead of early-returning on the stale `touchingRef`. On iOS the `setTimeout(0)` from `onScrollEndDrag`/`onTouchFinish` has usually already released, and the forced `setViewportInteractionActive(false)` is idempotent on an already-removed viewport — no over-release.

## Tests

- `hitTest.test.ts`: point (90,40) physically inside the viewport-owned block B resolves to A's end (offset 5) — correctly captures non-escape.
- `SelectionRoot.test.tsx`: the handle outside→in test asserts the range stays anchored to `outside#0` after dragging the end knob into the `inside` block's rect. The momentum test reproduces the Android sequence `touchStart → beginDrag → endDrag → momentumBegin` (no touchEnd) and asserts `[true] → [true, false]`. Both directions and both platform scenarios covered.

## Non-blocking nits

1. The three-state sentinel is well-documented inline; fine at this scale. If the scope dimension grows, a discriminated union (`{kind:'none'|'outside'|'viewport', id?}`) would read more explicitly than reusing `null` vs `undefined`.
2. The new hit-test case proves non-escape, but adding a sibling at y≈70 (inside C) resolving to C would also demonstrate that root-internal cross-block selection still works — i.e. the clamp is viewport-scoped, not a blanket freeze. Not required.
3. `onMomentumScrollBegin` dep array now includes `clearReleaseTimer`/`releaseIfIdle`; both are `useCallback`-stable, so no child-callback identity churn.

No blocking comments. Ship it.

PR #189, opened 2026-08-10
fix(conformance): close remaining harness blind spots

### LINKED ISSUE #106: Harden CommonMark conformance harness: regression-gated exits, PR trigger, comparison blind spots

kookyleo:
## Summary

The main regression gate, pull-request trigger, loud baseline mismatch, default-parser profile, and stale fix-report cleanup from #106 are already on `main`. This follow-up closes the remaining concrete blind spots:

- preserve all whitespace inside inline `<code>`, with a regression test that ordinary inline text still collapses whitespace
- classify severe proportional size collapse as review instead of pass, while keeping moderate scale-only changes as pass
- exclude locked generated fixtures from Prettier
- run harness unit tests in the conformance workflow and cache Cargo builds
- search up to 1,000 open issues during automatic deduplication
- generate both Bash and PowerShell single-case reproduction commands, and label platform-neutral README snippets as console commands

Closes #106

## Verification

- `pnpm test:unit` in `tests/markdown-conformance` — 3 pass, 0 fail
- `OFFICIAL_DIAGRAM_VISUAL_SELF_TEST=1 node tests/official-diagram-visual/scripts/tools/official-diagram-visual-workflow.mjs` — 12 checks pass
- rebuilt `supramark-markdown` and ran all 652 CommonMark semantic cases; the semantic-only run completed with the known 16 AST-target differences and loudly reported the expected baseline-target mismatch
- `bunx prettier --file-info tests/cases/_fixtures/commonmark/cases.json` — `ignored: true`
- JavaScript syntax checks — pass
- `bun run lint` — pass
- `bun run lint:types` — 0 findings
- `RUSTFLAGS="-C debuginfo=0" bun run quality` — pass
- `git diff --check` — pass

The explicit Rust flag only neutralizes this developer machines global `-fuse-ld=mold` setting; #186 carries the repository-level fix.

### DIFF
--- a/.github/workflows/markdown-conformance.yml
+++ b/.github/workflows/markdown-conformance.yml
@@ -132,6 +132,10 @@ jobs:
         working-directory: tests/markdown-conformance
         run: pnpm install --frozen-lockfile
 
+      - name: Run conformance harness unit tests
+        working-directory: tests/markdown-conformance
+        run: pnpm test:unit
+
       - name: Install Playwright Chromium
         if: env.RUN_VISUAL == 'true'
         working-directory: tests/markdown-conformance
@@ -139,6 +143,11 @@ jobs:
 
       - name: Set up Rust
         uses: dtolnay/rust-toolchain@1.97.0
+
+      - name: Cache Cargo build
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: markdown-conformance
 
       - name: Import locked data sources
         run: node tests/markdown-conformance/scripts/import.mjs "$SOURCE_NAME"
@@ -246,7 +255,7 @@ jobs:
             gh issue list \
               --repo "$GH_REPO" \
               --state open \
-              --limit 100 \
+              --limit 1000 \
               --json number,title,body \
               | jq -r \
                   --arg title "$issue_title" \
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,6 +4,7 @@ dist
 build
 .next
 .expo
+tests/cases/_fixtures
 
 # Vendored native primitives
 packages/renderers/rn-selection/native/selectable-rich-text
--- a/tests/markdown-conformance/README.md
+++ b/tests/markdown-conformance/README.md
@@ -31,7 +31,7 @@ tests/cases/_fixtures/commonmark/
 
 Run from the repository root:
 
-```powershell
+```console
 node tests/markdown-conformance/scripts/import.mjs commonmark
 node tests/markdown-conformance/scripts/validate.mjs commonmark
 ```
@@ -50,7 +50,7 @@ tests/cases/_fixtures/cmark-gfm/
   NOTICE.md
 ```
 
-```powershell
+```console
 node tests/markdown-conformance/scripts/import.mjs cmark-gfm
 node tests/markdown-conformance/scripts/validate.mjs cmark-gfm
 ```
@@ -59,7 +59,7 @@ node tests/markdown-conformance/scripts/validate.mjs cmark-gfm
 
 The micromark adapter discovers core Markdown-to-HTML assertions from `test/io/**/*.js` at the pinned commit. It excludes aggregation-only modules, stream tests, and constant-table tests. In an isolated capture environment it preserves Markdown input, expected HTML, upstream options, test names, and source line numbers, producing 1,151 normalized cases. Newly matching files require no changes to the import entry point or explicit path list.
 
-```powershell
+```console
 node tests/markdown-conformance/scripts/import.mjs micromark
 node tests/markdown-conformance/scripts/validate.mjs micromark
 ```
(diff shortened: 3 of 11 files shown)


mybee (approved):
Overall: a clean, tightly-scoped follow-up that closes the remaining blind spots from #106. Logic is correct, tests cover the new behavior, and no dead code is left behind. Approving with one clarification request and one optional cleanup, neither blocking.

## Per-change assessment

**1. `html-semantics.mjs` — preserve whitespace in inline `<code>`**
Correct fix. The original `parentTag === 'pre' || parentTag === 'code' && value.includes('\n')` was effectively `pre || (code && includes-newline)` due to precedence, so inline code without a newline got its whitespace collapsed. The new `pre || code` preserves all whitespace in `<code>`, matching CommonMark's literal-content semantics for inline code. `'code'` is not in `BLOCK_CONTAINERS`, so whitespace-only text nodes aren't accidentally emptied — good. Both new unit tests (preserve in code / collapse in ordinary inline) hit the right cases.

⚠ Clarification request: the PR body says you ran the 652 CommonMark semantic cases and "reported the expected baseline-target mismatch", but the description leans toward the ast-projection baseline. This `html-semantics` change only affects the **semantic** comparison path, so could you confirm the semantic pass/fail counts didn't shift unexpectedly? In theory this is stricter and may surface previously-masked whitespace diffs (a good thing) — just want to confirm no new false positives appeared.

**2. `official-diagram-visual-workflow.mjs` — downgrade severe proportional collapse to review**
Sound. Thresholds are `severeSizeDeltaThreshold=0.80`, `perceptualSimilarThreshold=0.08`. Dropping the `hasScaleOnlySizeDelta` guard means any severe area delta (≥0.80) with perceptual > 0.08 now escalates to review. The three self-tests are mutually consistent: proportional collapse area 0.958 → review, non-proportional 0.97 → review, moderate 0.36 → pass. `hasScaleOnlySizeDelta` is fully removed with no dangling references. An 80%+ area collapse is worth human eyes even when proportions hold.

**3. issue-report dual-platform repro commands**
Bash + PowerShell blocks both generate correctly. Extracting `representativeId` into a variable is a reasonable cleanup; the fallback chain is unchanged. The bash block uses `export`/`unset` and `$CASE_IDS` is valid in bash context. New unit test asserts both fences are present.

**4. CI workflow**
- `pnpm test:unit` runs before Playwright install — the script is added in the same PR's `package.json`, and `parse5` is available from the prior `pnpm install`. ✓
- `swatinem/rust-cache@v2` with `shared-key` — reasonable.
- `--limit 1000` for issue dedup — `gh issue list` paginates internally. ✓

**5. `.prettierignore` / README `console` fence / EOF newline**
All correct hygiene.

## Optional cleanup (non-blocking)

`hasSevereSizePerceptualMismatch` no longer reads `pixelBand`, but the production caller at line 864 and all 4 self-tests still pass `pixelBand: 'pass'`. The extra prop is harmlessly ignored, but since you're tidying up, dropping it from those call sites would be cleaner.

PR #198, opened 2026-08-12
feat(playground): unify Markdown feature routes

### LINKED ISSUE #197: feat(playground): consolidate the unified Markdown playground under /playground/*
## Problem

Supramark is a unified Markdown rendering library, but the current GitHub Pages routes split that product surface in the wrong way:

- `/preview/?feature=...` hosts the existing comprehensive Markdown and Feature preview with a Feature selector.
- `/playground/` hosts a separate D2-only application added by #145.

The comprehensive preview is the application that should represent Supramark in the root README. The current `/preview/` link also has a navigation bug: VitePress intercepts same-origin clicks as document routes, renders its 404 page, and only a full refresh reaches the static `preview/index.html` application.

This follows the README discoverability suggestion left after #139 was closed.

## Routing decision

Use one canonical, unified playground application:

- `/playground/` — the comprehensive Markdown and Feature playground with the selector visible.
- `/playground/core-markdown/`
- `/playground/gfm/`
- `/playground/math/`
- `/playground/mermaid/`
- `/playground/d2/`
- `/playground/dot/`
- `/playground/plantuml/`
- `/playground/echarts/`
- `/playground/vega-lite/`
- and equivalent routes for the remaining registered container and syntax features.

Changing the Feature selector must update the route, and opening a feature route directly must select that feature and its default example.

The existing `/preview/` application should move to this route family. The D2-only application must no longer own `/playground/`; retire it or fold any still-useful D2-specific behavior into the unified application. Existing `/preview/` URLs may remain as lightweight redirects to their canonical `/playground/` equivalents.

## GitHub Pages constraint

GitHub Pages does not provide a general SPA history fallback. Deep links must not depend on client-side routing alone. The build must emit a directly loadable document for every supported route, for example `playground/d2/index.html`, or use an equivalent static strategy that works on first navigation and browser refresh.

VitePress links into the standalone playground must also force a full document navigation instead of being intercepted as VitePress document routes.

## Scope

- Reuse and refactor `examples/react-web-csr` and its existing `FeaturePreview` registry rather than creating another engine-specific application.
- Move the comprehensive build output from `docs/public/preview` to `docs/public/playground`.
- Define stable URL slugs for every Feature in the selector.
- Keep the selector, editable Markdown, examples, loading state, responsive layout, and engine-specific rendering.
- Update the docs homepage iframe/button and all documentation links from `/preview/` to `/playground/`.
- Add a prominent root README link to the unified playground and explain that it runs without cloning or installing the repository.
- Remove or repurpose the duplicated `examples/d2-playground` build and workspace entry.

## Acceptance criteria

- [ ] `/supramark/playground/` opens the unified Markdown and Feature playground.
- [ ] `/supramark/playground/mermaid/` and `/supramark/playground/d2/` work on the first request and after refresh, without a VitePress 404.
- [ ] Equivalent deep links exist for every Feature exposed by the selector.
- [ ] Selecting a Feature updates the canonical path; browser back/forward restores the matching selection.
- [ ] The docs homepage button and iframe use `/playground/` and do not require a second refresh.
- [ ] The root README links the unified playground, not the D2-only page.
- [ ] D2 remains available at `/playground/d2/` through the same unified application.
- [ ] The standalone D2-only root application is removed or no longer deployed as the canonical playground.
- [ ] A deployed Pages smoke check confirms Markdown plus representative Mermaid, D2, PlantUML, ECharts, Vega-Lite, math, and container examples render.

Follow-up to #139, #145, and [the post-close README suggestion](https://github.com/Actrium/supramark/issues/139#issuecomment-5165842094).

kookyleo:
## Summary

- replace the retired `/preview/?feature=...` entry with one unified `/playground/`
- add stable static deep links for every registered feature, including `/playground/mermaid/`, `/playground/d2/`, `/playground/dot/`, and the remaining Markdown/container features
- make the feature selector update browser history and restore state on back/forward navigation
- remove the standalone D2-only playground and the unused generic CSR demo
- migrate docs, README links, GitHub Pages deployment, and visual-regression tooling to the new routes

Direct feature routes emit their own `index.html`, so GitHub Pages refreshes no longer depend on an SPA fallback. The old `/preview/` deployment output is removed without a redirect.

Closes #197

## Validation

- `bun run docs:build`
- `bun run quality`
- `bun run lint`
- `bun run lint:types`
- `bun run --cwd examples/react-web-csr test`
- `bunx tsc -p examples/react-web-csr/tsconfig.json --noEmit`
- `OFFICIAL_DIAGRAM_VISUAL_SELF_TEST=1 node tests/official-diagram-visual/scripts/tools/official-diagram-visual-workflow.mjs`
- headless browser checks for root/Mermaid/D2 direct loads, selector navigation, and back/forward state


boomsi (commented):
@kookyleo

**Review: verified hands-on — looks good to merge**

Verified the full route pipeline locally at 4e805d34 (built and browsed the real output, not just read the diff).

### Code review
- Slug map lives in exactly one place (`examples/react-web-csr/src/playground-routes.ts:3`), and `feature-registry.tsx:155` derives `routeSlug` via `routeSlugForFeature()` — an unregistered feature fails fast at module load instead of silently 404ing. Good trade.
- The static-deep-link approach is sound: `--base /supramark/playground/` makes asset refs absolute, so the `index.html` copies emitted by the `staticPlaygroundRoutes` plugin (`examples/react-web-csr/vite.config.ts:45`) need no rewriting — confirmed by inspecting the built `mermaid/index.html` against the root one.
- No stale references: repo-wide grep finds zero remaining `preview/?feature` links, `d2-playground` references, or `App.css` imports; `bun.lock` / `.eslintrc.js` / workflows cleaned up consistently. Generated docs (`docs/examples/*.md`) match the updated `doc-gen-example.ts`.

### 🟡 Non-blocking observations
1. In vite dev, manually typing a `/playground/<feature>/` URL silently falls back to `core-markdown`: dev `BASE_URL` is `/`, so the first path segment resolves to `playground` at `examples/react-web-csr/src/main.tsx:8`. Selector navigation produces `/mermaid/` and is unaffected — just worth knowing when debugging locally.
2. `tests/official-diagram-visual/scripts/tools/official-diagram-visual-workflow.mjs:500` inlines a copy of the `diagram-*` → slug map. Understandable since the .mjs can't import the TS module, but it will drift silently if a feature slug is ever renamed.
3. `FeaturePreview.tsx:218` passes `{ feature }` as pushState state, but the popstate handler re-resolves from the pathname and never reads `event.state` — harmless redundancy; either read the state or drop it.

### Tests
Ran locally:
- `bun test` (react-web-csr): **5 pass / 0 fail**
- `tsc -p examples/react-web-csr/tsconfig.json --noEmit`: clean
- `bun run build:wasm`: 4 wasm packages built
- `vite build --base /supramark/playground/`: **16 slug dirs emitted**, sub-page asset refs absolute and valid
- Headless Chrome against a local static server laid out as `/supramark/playground/` (Pages simulation): direct loads of `mermaid/`, `d2/`, `dot/`, `vega-lite/` each resolve the correct feature (checked `document.title`), root loads `core-markdown`, and the DOT page actually rendered a graphviz SVG — so the full engine path works from a deep link
- vite dev `/mermaid/` deep link: resolves correctly via SPA fallback
- `bun run lint` / `bun run quality` / `bun run docs:build`: all green; no `preview/` residue under `docs/public`

### Heads-up for maintainers (not blocking)
- `/preview/` is removed with no redirect — external links pointing at `actrium.github.io/supramark/preview/?feature=...` will 404 after merge. The PR description states this is deliberate; confirming it's a conscious product call.
- Suggest a post-deploy click-through of the README playground link plus a deep-link refresh on the real Pages output, and one full dispatch of the visual-regression workflow to confirm selector interaction on the new routes.

Ready to merge.

boomsi (approved):
Verified hands-on — details in the review comment above. All local checks green (tests, tsc, build, headless deep-link checks, lint, quality, docs build).

PR #200, opened 2026-08-13
fix(rn): render Markdown images in stable containers

### LINKED ISSUE #195: [renderers/rn] Markdown images render as a "[Image: alt]" text placeholder instead of actual images
## Summary

In the React Native renderer, image syntax `alt` renders as a text placeholder `[Image: alt]` instead of an actual image. The RN branch is an intentional stub — the code comment says *"could use the Image component in the future"* — so this tracks implementing real image rendering on RN.

## Reproduction

Input:

```md
![图片](http://politics.people.com.cn/NMediaFile/2026/0811/MAIN1786411359243KHAEDRCEOA.jpg)
```

- **Actual (RN):** `[Image: 图片]`
- **Expected:** a rendered image. The web renderer already does this correctly.

## Root cause

`packages/renderers/rn/src/Supramark.tsx`, `renderInlineNode` → `case 'image':` (L1038–1046):

```tsx
case 'image': {
  const imageNode = node;
  // Show images as text for now in RN (could use the Image component in the future)
  return (
    <Text key={key} style={styles.imageText}>
      [Image: {imageNode.alt || imageNode.url}]
    </Text>
  );
}
```

Supporting evidence it was never implemented:

- The `react-native` import at `Supramark.tsx:2` does not include `Image` (`Text, View, Linking, TouchableOpacity, Dimensions`).
- A dedicated placeholder style `imageText` exists for this stub at `packages/renderers/rn/src/styles.ts:268-271`.
- The web renderer renders the same node correctly as `` at `packages/renderers/web/src/Supramark.tsx:1575-1586`, so the AST node type (`'image'`) and parsing are fine — only the RN layer is stubbed.

The placeholder shows `[Image: 图片]` (the alt text) rather than the URL because the expression is `alt || url`.

## Why it isn't a one-line fix

1. **Sizing.** Markdown images carry no dimensions, and RN `Image` needs a size constraint. The usual approach is `onLoad` → read `nativeEvent.source.width/height` → set `aspectRatio` with `width: '100%'`, which requires component state — so `case 'image':` should delegate to a small `MarkdownImage` component rather than inline the hook.

2. **Block vs. inline.** `paragraph` wraps its children in `<Text>` (`Supramark.tsx:446-451`), so an image rendered inside becomes an *inline* embedded image (emoji-like, fixed pixel size) rather than a block image that fills the width. Standalone images (e.g. a news photo on its own line) want block behavior, which likely means `paragraph` lifting an image-only run out of the `<Text>` wrapper.

Edge cases to decide on: image mixed with inline text (`text img text`) and image-as-link (`[img](link)`).

## Environment

- Renderer: `@supramark/rn`
- Branch: `main`

boomsi:
## Summary

- render Markdown image nodes with React Native's `Image` instead of the `[Image: alt]` placeholder
- reserve a stable 200×200 container for standalone images so virtualized-list row heights do not change after loading
- default block and inline images to `cover` while allowing hosts to override container dimensions and `resizeMode` through `SupramarkStyles`
- group consecutive image-only paragraphs into one native horizontally scrolling row with an 8px gap
- preserve platform-native dragging, momentum, nested scrolling, and directional locking instead of implementing gallery movement with a JavaScript `PanResponder`
- constrain the gallery viewport to the configured image height so flex layouts cannot create blank vertical space
- round block-image containers by 8px by default while allowing hosts to override both gallery spacing and corner radius
- preserve inline image flow and standalone image-link interactions

## Root cause

The React Native renderer intentionally left the `image` branch as a text placeholder. A direct replacement inside paragraph `Text` would not support block sizing, so image-only paragraphs and image-only links now use a dedicated stable block container while mixed-content images remain inline.

The image gallery uses React Native's native horizontal `ScrollView` so touch arbitration remains in the platform scrolling system. Its viewport is constrained to the configured image height, while `directionalLockEnabled` and `nestedScrollEnabled` preserve native directional and nested-scroll behavior. No host gesture or host-specific dependency is exposed through the Supramark API. Web behavior is unchanged.

## Host configuration

```tsx
<Supramark
  markdown={markdown}
  styles={{
    imageContainer: { width: 320, height: 180 },
    imageGallery: { gap: 12 },
    image: { resizeMode: 'contain' },
  }}
/>
```

## Test plan

- [x] `bun --filter @supramark/rn test` — 130 passed, 0 failed
- [x] `bun test packages/renderers/rn/__tests__/image.test.tsx` — 8 passed, including native-scroll and viewport-height assertions
- [x] `bun --filter @supramark/rn build`
- [x] targeted ESLint and Prettier checks
- [x] `git diff --check`

Fixes #195


kookyleo (approved):
Re-reviewed at 62a07b87. The four previous should-fix items are actually fixed: list images reach the stable block container, hard break matches soft break, disabled footnotes no longer nest a View under Text, and empty/relative URLs render a placeholder. @supramark/rn tests: 144 pass / 0 fail (image.test.tsx 22/22).

Nits (non-blocking): mixed-content inline images still have no onError fallback; TouchableOpacity around tappable images does not copy accessibilityLabel; failed state is not reset on URL change.

PR #201, opened 2026-08-13
fix(rn): render-cache review follow-ups (maxBytes:0, freeze timing)

mybee:
## Summary
Follow-up to #160 (render-cache follow-ups, merged). Two small review-driven fixes:

- **`maxBytes: 0` symmetry.** `resolveRendererCachePolicy` now distinguishes "not configured" (→ 8 MB default) from an explicit `0` / non-positive / non-finite value (→ `undefined`, no byte cap). Pre-fix, an explicit `0` was indistinguishable from "unset" and silently fell through to the 8 MB default, so a host could not opt out of the byte bound. Mirrors how `ttl` and `LRUCache.reconfigure` already resolve.
- **Freeze timing.** Moved the dev-only `deepFreezeAst` into the `buildParsedDocument` factory so the parsed snapshot is frozen *before* it is handed to anything that might retain it, instead of at the `setParsedDocument` call site.

## Test plan
- [x] New `resolveRendererCachePolicy maxBytes resolution` describe block (3 cases): `maxBytes:0` → `undefined`; negative / `Infinity` / `NaN` → `undefined`; positive retained + override-wins + unset → default.
- [x] Existing `renderCache` suite green (28/28).
- [x] `tsc --noEmit` clean on `@supramark/rn`.

### DIFF
--- a/packages/renderers/rn/__tests__/renderCache.test.ts
+++ b/packages/renderers/rn/__tests__/renderCache.test.ts
@@ -256,3 +256,29 @@ describe('getRendererCache policy reconfiguration (#124 #3)', () => {
     expect(after).not.toBe(before); // new instance after clear
   });
 });
+
+describe('resolveRendererCachePolicy maxBytes resolution (#124 follow-up)', () => {
+  // Pre-fix, an explicit maxBytes: 0 fell through to the 8 MB default
+  // (indistinguishable from "unset"), so a host could not opt out of the byte
+  // cap. Now 0 (or any non-positive / non-finite value) means "no byte bound",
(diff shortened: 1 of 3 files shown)


kookyleo (approved):
Reviewed. maxBytes: 0 maps to no byte cap (undefined), not a disabled cache and not a zero-capacity cache that still allocates. Freeze now runs inside the factory before getOrCreate stores the AST, which closes the unfrozen-snapshot window. Tests for the 0 / non-finite / override-wins cases passed.

Nits (non-blocking): document that maxBytes:0 is not a kill switch (maxSize:0 is); no Object.isFrozen assertion.

PR #208, opened 2026-08-20
feat(wikilink): add [[target]] parsing support across the full stack

### LINKED ISSUE #185: Add WikiLink parsing support to supramark-markdown
## Motivation

Knowledge-base Markdown produced by tools such as Obsidian and Logseq commonly uses WikiLinks. The current `supramark-markdown` parser has no dedicated WikiLink node or feature, so expressions such as `[[Project Plan]]` are preserved as ordinary text and downstream renderers cannot distinguish or resolve them safely.

## Proposed scope

Add parser-level support for the common forms:

- `[[target]]`
- `[[target|label]]`
- optionally, heading fragments such as `[[target#section]]` if they fit the initial scope

The AST should preserve the target, optional display label, and the existing UTF-8/UTF-16 source-position contract. Resolution to a file or URL should remain a downstream concern because it depends on workspace rules.

## Expected behavior

- WikiLinks are represented by a dedicated serializable AST node (or another unambiguous structured representation).
- Escaped forms and WikiLink-looking text inside code spans/fences remain literal.
- Malformed or unclosed forms degrade predictably and retain source positions/diagnostics.
- Existing CommonMark/GFM behavior remains unchanged when the syntax is not used (a feature gate would also be reasonable).

## Downstream use case

Markon can use the structured node to resolve WikiLinks within the active workspace while applying its existing path and capability checks.

mybee:
## Summary

Closes #185 — full vertical slice for knowledge-base Markdown (Obsidian/Logseq) WikiLink syntax.

- **Rust**: `WikiLinkScanner` inline rule registered `.before::<LinkScanner>` so wiki semantics win over CommonMark reference definitions. Gated by a runtime `ParseOptions.wikilink` flag (default **off** — all existing conformance baselines stay byte-identical). `check()` returns `None` by design so `parse_link_label`'s single-char bracket counting in skip mode is not disturbed (`[text [[x]]](url)` keeps its CommonMark link semantics). Syntax: `[[` to first `]]`; first `|` splits label, first `#` in target part splits section; `[[#section]]` allowed; malformed/unclosed forms degrade to literal text with **no diagnostic**.
- **WASM**: `parse_json_with_options` plumbing via manual `Reflect::get` (no new deps).
- **Core TS**: `SupramarkWikiLinkNode`, `wikilink` parse option, loader/adapter contracts, `expandOpaqueContainers` option threading. Feature enabled in config turns the parser flag on; explicit `{ wikilink: true }` overrides. RN module lacking option support throws an explicit error rather than silently degrading.
- **Feature package** `@supramark/feature-wikilink` with a `resolveWikiLink` callback (resolution stays downstream — Markon resolves within workspace).
- **Web/RN renderers**: styled non-navigable text when resolver is absent/null (no fake `href="#"`); literal source rebuild when feature disabled.
- **Docs**: ast-spec §6.11, feature doc, manifest entry, example registry + demo config.

## Test plan

- [x] `cargo test -p supramark-markdown` — 140 tests (9 new wikilink tests: forms, positions, degradation, reference-definition precedence, escapes, code spans)
- [x] `cargo build -p supramark-markdown` + `bun run build:wasm markdown`
- [x] `bun run test` — all workspace packages pass (core 26 incl. 4 new wikilink; feature-wikilink 5; rn 122; web)
- [x] `bun run lint` (eslint + features:lint 22/22 + cjk check) clean
- [x] `bun run features:sync` — regenerated docs/bundles reviewed
- [x] Conformance `run-visual.mjs` — **PASS, zero regressions** across `commonmark` / `cmark-gfm` / `micromark` (default-off flag ⇒ no baseline diffs)

## Notes

- `examples/react-web-csr/src/all-features.ts` and `examples/react-native/src/all-features.ts` are regenerated by `features:sync` but intentionally left untracked (they were removed from the tree in 99d3ffce).
- RN native FFI for the parser option is **not** in this PR (per scope decision); RN throws an explicit error when wikilink is requested but the module lacks option support. A follow-up issue should track native FFI / TurboModule bridging.
- `ParseOptions.wikilink` is a runtime flag, not a cargo feature — default-off keeps build profiles stable and minimizes merge conflict surface with the unmerged `fix/issue-184-conformance-options` branch.


kookyleo (changes requested):
DO NOT MERGE — the Rust parser is solid; the host-facing config/demo is not. This PR now also conflicts with `main` (playground files vs #198 if it lands, and `packages/renderers/rn/src/Supramark.tsx` vs merged #200 / #201). Rebase after those, and do not copy the broken demo into the unified playground.

`SupramarkConfig.features` is `FeatureConfig[]` (`{ id, enabled, options }`). `isFeatureEnabled` / `getFeatureOptionsAs` key on `f.id`. The playground and the feature README pass `features: [wikilinkFeature]` (id is on `metadata.id`) plus a non-existent `featureConfigs` field. I ran `parse()` with that object: no `wiki_link` nodes, resolver never read. The working shape is `features: [createWikilinkFeatureConfig(true, { resolveWikiLink })]`, same as math. Please fix `FeaturePreview`, `README.zh.md`, and the documentation example, and add a core test that enables the flag via config rather than `{ wikilink: true }`. CI is green because every automated parse uses the explicit flag.

Also:

1. Inline math is a text post-pass. With the option on, `$[[foo]]$` is no longer `math_inline` — it becomes `$` + wiki_link + `$`. Code spans are protected; math is not. Add a parser test and either scan math before WikiLink or skip `[[` inside `$...$`.
2. RN `documentCacheKey` still omits config with a comment that parse ignores it. This PR makes that false whenever cache is on.
3. Enabling the feature on RN currently throws on every `parse()` (no `parseJsonWithOptions` on the native ABI). Do not advertise a working RN path until that exists.
4. If #198 lands first, `routeSlugForFeature()` at registry init will throw on `wikilink` unless `{ feature: 'wikilink', slug: 'wikilink' }` is added in the same change.

Default web/RN path is not an automatic XSS: `href` comes only from `resolveWikiLink`. A naive resolver `({ target }) => target` would turn `[[javascript:alert(1)]]` into a live link; CommonMark `[js](javascript:alert(1))` never becomes a link. Please document that hosts must sanitize resolver output.

PR #209, opened 2026-08-20
fix(math): parse inline math containing cmark-escaped braces (#207)

### LINKED ISSUE #207: fix(math): render inline and block formulas using \\widehat instead of showing raw TeX
## Problem

Math expressions containing a hat accent are not rendered in the Markdown preview. I tested both common spellings:

- `\widehat\rho`
- `\widehat{\rho}`

Both fail in both block and inline contexts. The preview shows the literal dollar signs/backslashes instead of rendered math.

## Minimal reproduction

```markdown
Block form:

$$
\widehat{\rho}_{\Gamma,q}(a,s) \in \{0, ?, 1\}
$$

Inline form: $\widehat{\rho}=1$
```

The same result is obtained when the source uses `\widehat\rho` instead of `\widehat{\rho}`.

## Actual result

The preview contains literal text such as:

```
$\widehat{\rho}=1$
\widehat{\rho}_{\Gamma,q}(a,s)
```

rather than typeset math. This was observed in the Supramark preview used by Markon.

## Expected result

Both block and inline expressions should either:

1. be passed to the configured math engine and rendered; or
2. produce a clear, actionable unsupported-macro diagnostic.

They should not silently fall back to raw TeX source.

## Notes

Other simple mathematical expressions appear to render, so this may be one of:

- incomplete support for the `\\widehat` command;
- different handling of inline and block math;
- escaping/tokenization before the math engine;
- math support not being enabled for this renderer.

Please add a regression test covering both delimiter forms and both spellings, and document the supported math dialect/macros.

mybee:
## Summary

- Fix #207: inline math containing cmark backslash-escaped punctuation (`\\{`, `\\}`, `\\$`, …) collapsed to literal text instead of rendering. Example: `$\\widehat{\\rho}_{\\Gamma,q}(a,s) \\in \\{0, ?, 1\\}$`.
- Root cause: cmark decodes `\\{`→`{` before the inline math delimiter scan runs. The decoded `value` is then shorter than the source span, the `$` parity check miscounts, and the whole inline math falls back to a plain text node. Block `$$…$$` was unaffected (line-based block rule, not an inline scan).
- Fix: thread the raw `source` slice through `AstV2Ctx` → `map_inline_text`. Fast path (`value.len() == source span`) scans value directly. Slow path builds a `RawValueMap` by lock-stepping raw↔value bytes, scans the raw slice for `$` / `[^…]` delimiters, and maps offsets back to value for text/footnote content. Math `value` is sliced from raw so `\\{` reaches the TeX engine unchanged. Variable-width decodings (HTML entities) make `RawValueMap::new` return `None` → fall back to the pre-fix plain text node (no regression).
- Refactor: extract the CommonMark escapable punctuation set into `pub fn is_escapable_punct(char)` in `escape.rs` and reuse it from the AST v2 inline mapper, removing a byte-level duplicate.

## Test plan

- [x] `cargo test --package supramark-markdown` — 46 passed (incl. `public_api_inline_math_widehat_escaped_braces_207`)
- [x] `bun test packages/core/__tests__/parser.test.ts` — 24 passed (incl. `#207` inline + block regression cases)
- [x] `bun --filter @supramark/feature-math test` — 3 passed
- [x] wasm rebuilt (`bun run build:wasm markdown`)

### DIFF
--- a/crates/supramark-markdown/src/plugins/cmark/inline/escape.rs
+++ b/crates/supramark-markdown/src/plugins/cmark/inline/escape.rs
@@ -8,6 +8,21 @@ use crate::parser::inline::{InlineRule, InlineState, TextSpecial};
 use crate::plugins::cmark::inline::newline::Hardbreak;
 use crate::{MarkdownParser, Node};
 
+/// Whether `chr` is a CommonMark backslash-escapable punctuation character.
+///
+/// This is the single source of truth for the escapable set
+/// (<https://spec.commonmark.org/0.30/#backslash-escapes>). Other modules
+/// (e.g. the AST v2 inline mapper) reuse it instead of duplicating the set.
+pub fn is_escapable_punct(chr: char) -> bool {
+    matches!(
+        chr,
+        '\\' | '!' | '"' | '#' | '$' | '%' | '&' | '\'' | '(' | ')'
+            | '*' | '+' | ',' | '-' | '.' | '/' | ':' | ';' | '<' | '='
+            | '>' | '?' | '@' | '[' | ']' | '^' | '_' | '`' | '{' | '|'
+            | '}' | '~'
+    )
+}
+
 pub fn add(md: &mut MarkdownParser) {
     md.inline.add_rule::<EscapeScanner>();
 }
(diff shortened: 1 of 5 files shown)


kookyleo (approved):
Reviewed. The #207 slow path is the right fix: cmark decodes \{ before the inline math scan, and keeping the raw slice in the math value is what TeX needs. is_escapable_punct matches CommonMark 0.30 (32 ASCII punctuation characters). cargo test -p supramark-markdown passed on the PR head, including the #207 cases.

Nits (non-blocking): first-pass scanning still breaks $\{1\}$ and $\{2\}$ in one paragraph; tests only cover a single math run.

PR #210, opened 2026-08-20
fix(rn): correct d2 v0.7.1 nested svg outer viewBox to fit inner content

boomsi:
## Bug

d2 v0.7.1 outputs a **nested svg** structure (an outer `<svg>` wrapping an inner one). The outer `<svg>`'s `viewBox` dimensions are wrong — much smaller than the inner content's actual dimensions. As a result:

- Only the top-left corner of the diagram is visible
- Most text falls outside the viewport (frame renders, text doesn't)
- Affected: any host using d2 v0.7.x with the React Native renderer

This regression doesn't happen with d2 v0.6 or mermaid output.

## Fix

Extracted into `fixD2NestedViewBox()` in `packages/renderers/rn/src/svgUtils.ts`, with 15 new unit tests in `svgUtils.test.ts`. `DiagramNode` delegates to it; when the outer and inner viewBox dimensions disagree by more than **2x**, the fix:

1. Replaces the outer `viewBox` with the inner content dimensions
2. Returns the new intrinsic width/height for `DiagramNode` to recompute `chartWidth` and `height`

The **2x threshold** is a guard rail: d2 v0.7.1 produces a ~28x mismatch, so 2x leaves a 14x safety margin against d2 v0.6 / mermaid output that legitimately have a different outer viewBox without being buggy. The threshold is exported as `D2_NESTED_VIEWBOX_RATIO_THRESHOLD` for testability.

## Long-term

Track upstream d2 for a fix. This is a workaround for the d2 v0.7.1 SVG output bug.

## Risk

- **Renderer-only** change; no API surface change
- **Cross-platform** (web / iOS / Android / Windows all affected equally)
- Triggers only when d2 outputs nested svg AND outer/inner disagree by >2x — false positive risk is low
- Threshold is exported; tests cover boundary cases (exactly 2x → not applied, 2.01x → applied)

## Test plan

- [x] **15 unit tests** in `svgUtils.test.ts > fixD2NestedViewBox`:
  - d2 v0.7.1 actual case (28x mismatch) → fix applied, outer viewBox replaced
  - d2 v0.6 case (matching dimensions) → not modified
  - mermaid (single svg) → not modified
  - Threshold boundary cases (exactly 2x → not modified, 2.01x → modified)
  - Malformed viewBox (fewer than 4 components, non-numeric width/height, zero dimensions)
  - Outer dimensions zero (avoids division by zero)
  - Only one svg with viewBox → not modified
  - Three-level nested svgs (documents current 2-level limit)
- [ ] Manual render: d2 v0.7.1 with text labels — expect text visible inside the frame
- [ ] Manual render: d2 v0.6, mermaid, single-svg output — expect no behavior change

### DIFF
--- a/packages/renderers/rn/__tests__/svgUtils.test.ts
+++ b/packages/renderers/rn/__tests__/svgUtils.test.ts
@@ -487,3 +487,162 @@ describe('stripRootSvgSize', () => {
     expect(result).not.toContain('height=');
   });
 });
+
+// ============================================================================
+// fixD2NestedViewBox — d2 v0.7.1 nested-svg outer-viewBox bug
+// ============================================================================
+
+import {
+  fixD2NestedViewBox,
+  D2_NESTED_VIEWBOX_RATIO_THRESHOLD,
+} from '../src/svgUtils';
+
+describe('fixD2NestedViewBox', () => {
+  // ----- d2 v0.7.1 nested case (the actual bug) -----
+
+  test('d2 v0.7.1: replaces the outer viewBox with the inner content dimensions', () => {
+    // Outer viewBox is "0 0 10 10" but the inner d2-svg is actually 280x250.
+    // Real-world ratio is ~28x, well above the 2x threshold.
+    const svg =
+      '<svg class="d2-wrapper" viewBox="0 0 10 10">' +
+      '<svg class="d2-svg" viewBox="-1 -1 280 250">' +
+      '<g/></svg></svg>';
+    const result = fixD2NestedViewBox(svg, 10, 10);
+    expect(result.applied).toBe(true);
+    expect(result.intrinsicWidth).toBe(280);
+    expect(result.intrinsicHeight).toBe(250);
+    // Outer viewBox is the FIRST match → it's the one replaced.
+    expect(result.svg).toMatch(/<svg class="d2-wrapper" viewBox="0 0 280 250">/);
+    // The inner d2-svg is preserved.
+    expect(result.svg).toContain(
+      '<svg class="d2-svg" viewBox="-1 -1 280 250">'
+    );
+  });
+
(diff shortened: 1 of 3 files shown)


kookyleo (changes requested):
Requesting changes — the helper is wired (HEAD replaced the inline copy, replacement is anchored to the outer `<svg viewBox>` tag), and unit tests of that helper pass. I cannot find the bug this is fixing in this tree.

d2-little v0.7.1 nested output is constructed as outer `viewBox="0 0 W H"` wrapping inner `viewBox="{left} {top} W H"` with the **same W/H** (`crates/d2-little/src/svg_render/mod.rs`). I scanned 646 nested-svg goldens: every outer/inner size ratio is 1.000×. Calling `fixD2NestedViewBox` on `elk_golden/issue34.svg`, `japanese-basic/dagre/sketch.exp.svg`, and `d2-v2-container-nested-grid.svg` all return `applied: false`.

The “~28× real-world” case is the synthetic `viewBox="0 0 10 10"` vs inner `280×250` (280/10 = 28). Nothing in-repo emits `d2-wrapper` or a 10×10 outer around a 280-wide `d2-svg`.

The crop you describe (frame visible, text clipped) is already documented on `stripRootSvgSize`: Android `react-native-svg` plus an inner viewBox origin offset (`-1 -1` / `11 11`) if inner width/height are stripped. This workaround never rewrites that inner origin, and on real payloads it never rewrites the outer either.

Please either:

1. Check in the actual SVG that cropped on device and a `DiagramNode` test that asserts `SvgXml` xml/size after the helper, or
2. Drop the outer-viewBox rewrite and fix the inner origin / RN-svg nested path instead.

Also gate on `engine === 'd2'` (or inner `class` containing `d2-svg`). This currently runs for every diagram.

#211 (`diagram.maxWidth`) is already on `main` and auto-merges with this file. Rebase if you keep the helper.

PR #211, opened 2026-08-20
feat(rn): add maxWidth to SupramarkDiagramConfig for host-driven chart cap

boomsi:
## Motivation

Hosts whose diagram container is narrower than the window (e.g. a desktop chat bubble with a fixed `maxWidth`) currently cannot prevent the diagram from overflowing. The renderer uses `windowWidth × 0.9` as the chart width cap, ignoring the container's own constraints — so a narrow bubble gets a wide diagram that overflows the bubble's right edge.

## Change

Add an optional `maxWidth?: number` field to `SupramarkDiagramConfig`. The renderer now clamps chart display width to:

```
min(windowWidth × 0.9, host.maxWidth ?? Infinity)
```

Unset keeps the existing screen-derived cap. A type-guard (`typeof === 'number' && > 0`) rejects 0 / negative / NaN so junk callers don't change behavior.

## Why renderer, not core

The clamp belongs in the renderer because it's about the display surface, not about diagram semantics. Adding it to the config type is the lightest possible API surface change (one optional field).

## Cross-platform

This change benefits **every** host (iOS / Android / macOS / Windows / web), not just Windows. The motivation is from a desktop chat bubble, but the fix is universal.

## Risk

- Pure additive API: `maxWidth` is optional, default behavior unchanged
- 2 files, 18 lines
- No new types or re-exports

## Test plan

- [ ] `maxWidth` unset → behavior identical to before
- [ ] `maxWidth: 300` on a 400dp-wide container → chart capped at 300dp
- [ ] `maxWidth: 0` or negative → ignored, falls back to screen cap
- [ ] `maxWidth` larger than `windowWidth × 0.9` → screen cap wins

### DIFF
--- a/packages/core/src/ast.ts
+++ b/packages/core/src/ast.ts
@@ -205,6 +205,15 @@ export interface SupramarkDiagramConfig {
   /** Default timeout (milliseconds), used for engines without their own config */
   defaultTimeoutMs?: number;
 
+  /**
+   * Host-provided cap on diagram display width (dp). Diagrams clamp their
+   * display width to min(windowWidth × 0.9, maxWidth). Hosts whose container
+   * is narrower than the window (e.g. a desktop chat bubble with a fixed
+   * maxWidth) pass their own cap so the diagram doesn't overflow the
+   * container. Unset means the screen-derived cap applies alone.
+   */
+  maxWidth?: number;
+
   /** Default cache configuration */
   defaultCache?: {
     enabled?: boolean;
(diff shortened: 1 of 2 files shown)


kookyleo (approved):
Reviewed. diagram.maxWidth is not dead: config.diagram already flows into DiagramNode, and maxChartWidth is min(window*0.9, maxWidth). 0 / negative / NaN behave as unset. Layout-only, correctly omitted from the render cache key.

Nits (non-blocking): units are RN dp; the 0.6 floor remains; web ignores the field; no test.

PR #212, opened 2026-08-20
fix(build): graphviz Windows build fix for Chinese code page / MSYS2

boomsi:
## Bug

On Chinese Windows (system code page 936 / GBK), MSVC parses source files using the local code page by default. Graphviz's UTF-8 source files trigger C4819 warnings that cascade into bizarre syntax errors — e.g. `htmltable.c` reporting an "undeclared identifier" for `i`, because the multi-byte sequence gets mis-read and eats the surrounding quotes/semicolons.

A second, independent failure mode: Git Bash (MSYS2) auto-converts the `/utf-8` value in `-DCMAKE_C_FLAGS=/utf-8` to a Unix path, polluting CMakeCache (`C:/Program Files/Git/utf-8`).

## Fix

In `crates/graphviz-anywhere/scripts/build-windows.sh`, on the cmake invocation:

- Add `-DCMAKE_C_FLAGS=/utf-8` and `-DCMAKE_CXX_FLAGS=/utf-8` so MSVC parses sources as UTF-8
- Wrap with `MSYS2_ARG_CONV_EXCL='-DCMAKE_C_FLAGS;-DCMAKE_CXX_FLAGS'` so Git Bash's path conversion skips exactly these two arguments

The change is scoped to this script's cmake call — `common.sh` is untouched, so cross-platform shared parameters are unchanged.

## Risk

- Windows-only
- 1 file, 12 lines
- No effect on non-Windows build paths
- Both args are already optional; a non-Chinese, non-MSYS2 environment sees the same flags added but with no observable difference (the values are still `/utf-8`, which is a valid MSVC flag)

## Test plan

- [ ] Run `./build-windows.sh` on a Chinese Windows (code page 936) — expect Graphviz C4819 errors gone, full build succeeds
- [ ] Run on English Windows — expect no regression
- [ ] Confirm `common.sh` is unchanged (`git diff main...HEAD -- common.sh` should be empty)

### DIFF
--- a/crates/graphviz-anywhere/scripts/build-windows.sh
+++ b/crates/graphviz-anywhere/scripts/build-windows.sh
@@ -126,10 +126,19 @@ prepare_graphviz_source "${GV_PATCHED}"
 # htmllex.c fails with "Cannot open include file: 'expat.h'".)
 log_info "Configuring Graphviz..."
 mkdir -p "${BUILD_DIR}/graphviz"
+# MSVC 源码编码：中文 Windows（代码页 936/GBK）下 MSVC 默认按本地代码页解析源文件，
+# graphviz 源码/头文件中的 UTF-8 字符会触发 C4819 并连锁产生语法错（如 htmltable.c
+# 的「i 未声明」假错——多字节序列错读吞掉引号/分号）。强制 /utf-8 解析。
+# 仅加在 Windows 生成器调用上，不动 common.sh 的跨平台共享参数。
+# MSYS2_ARG_CONV_EXCL：Git Bash 会把值里的 /utf-8 当 Unix 路径转换成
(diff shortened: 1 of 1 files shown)


kookyleo (approved):
Reviewed and compiled on Windows MSVC (VS 2022 BuildTools, cl 19.44). After git submodule update --init of crates/graphviz-anywhere/graphviz, scripts/build-windows.sh --arch x86_64 completed (GRAPHVIZ_RC=0). /utf-8 + MSYS2_ARG_CONV_EXCL is the right MSVC source-charset fix.

Nits (non-blocking): CMAKE_C_FLAGS=/utf-8 replaces rather than appends; PR CI does not run this script (release-asset smoke only).

PR #213, opened 2026-08-20
feat(windows): markdown-native-rn React Native Windows port

boomsi:
## Motivation

`@supramark/markdown-native-rn` currently has native bindings for iOS / Android / macOS only. On Windows, the package's `src/index.ts` `Platform.select` has no `windows` branch, so `parse()` falls back to the wasm path (`@supramark/markdown-web`) — which has not been verified to work in Hermes (the JS engine used by react-native-windows).

## Change

Add Windows native bindings for the markdown parser, following the same pattern as the existing iOS / Android / macOS bindings:

**New C++/WinRT module** — `crates/supramark-markdown/packages/react-native/windows/SupramarkMarkdownNative/`:
- `ReactPackageProvider.cpp` / `.h`: registers the module with the RNW runtime
- `SupramarkMarkdownModule.h`: defines the FFI signature, `dllimport` function declarations, and the `ReactPackage` template

**JS layer**:
- `package.json`: `files` adds `windows`, `peerDependenciesMeta` adds `react-native-windows`, adds a `react-native-windows` platform extension
- `src/index.ts`: `Platform.select` adds a `windows` branch
- `src/resolveNative.ts`: adds the windows resolution path
- `scripts/prepare-native.js`: Windows binary staging (copy `.dll` / `.h` / `.lib` to `windows/Frameworks/`)

**Build script** — `crates/supramark-markdown/scripts/build-windows.sh` (154 lines): cross-compiles the Rust crate to a Windows `.dll` via the C ABI.

**`gitignore`**:
- Root `.gitignore`: ignore Windows build artifacts
- `crates/supramark-markdown/packages/react-native/.gitignore`: add `windows/Frameworks/` (consistent with existing `ios/Frameworks/`, `macos/Frameworks/`)

## Risk

- Windows-only — no impact on iOS / Android / macOS
- 9 files, 339 lines
- FFI signature must match the Rust crate's `extern "C"` exports (verified by `build-windows.sh` linking against `output/windows-x86_64/include/supramark_markdown.h`)
- `prepare-native.js` Windows branch handles `x86_64` and `arm64` architectures

## Test plan

- [ ] Run `./build-windows.sh` in `crates/supramark-markdown` — expect `output/windows-x86_64/bin/supramark_markdown_native.dll` produced
- [ ] Run `npm run prepare-native` — expect `windows/Frameworks/{bin/*.dll, include/*.h, lib/*.lib}` populated
[...]

PR #215, opened 2026-08-20
feat(windows): d2 diagram engine React Native Windows port

boomsi:
## Motivation

`@actrium/supramark-d2-native-rn` currently has native bindings for iOS / Android / macOS only. On Windows, the `createReactNativeDiagramEngine` lookup fails to find a Windows adapter and falls back to the wasm loader — which has not been verified to work in Hermes.

## Change

Add Windows native bindings for the d2 engine, following the same pattern as the iOS / Android / macOS bindings (and parallel to the `markdown-native-rn` and other engine PRs):

**New C++/WinRT module** — `crates/d2-little/packages/react-native/windows/SupramarkD2Native/`:
- `ReactPackageProvider.cpp` / `.h`: registers the module with the RNW runtime
- `SupramarkD2Module.h`: defines the FFI signature, `dllimport` function declarations, and the `ReactPackage` template

**JS layer**:
- `package.json`: `files` adds `windows`, `peerDependenciesMeta` adds `react-native-windows`, adds a `react-native-windows` platform extension
- `src/index.ts`: `Platform.select` adds a `windows` branch
- `scripts/prepare-native.js`: Windows binary staging (copy `.dll` / `.h` / `.lib` to `windows/Frameworks/`)

**`gitignore`**: `windows/Frameworks/` added (consistent with existing `ios/Frameworks/`, `macos/Frameworks/`).

**Build script** — `crates/d2-little/scripts/build-windows.sh` (121 lines): cross-compiles the Rust crate to a Windows `.dll` via the C ABI.

## Risk

- Windows-only — no impact on iOS / Android / macOS
- 8 files, 297 lines
- FFI signature must match the Rust crate's `extern "C"` exports
- `prepare-native.js` Windows branch handles `x86_64` and `arm64` architectures
- `build-windows.sh` is the most likely to need iteration on a real Windows machine (CMake / MSVC toolchain availability, Hermes quirks)

## Test plan

- [ ] Run `./build-windows.sh` in `crates/d2-little` — expect `output/windows-x86_64/bin/supramark_d2_native.dll` produced
- [ ] Run `npm run prepare-native` — expect `windows/Frameworks/{bin/*.dll, include/*.h, lib/*.lib}` populated
- [ ] In a host RNW app, side-effect `import '@actrium/supramark-d2-native-rn'` at entry top — expect `createReactNativeDiagramEngine()` resolves to the native adapter
- [ ] Render a sample d2 diagram — expect SVG output identical to other platforms

## Pattern

This PR is the second of three parallel engine ports. The `mermaid` port is already in PR #214; `plantuml` follows in the next PR. Identical structure, separate reviews.

### DIFF
--- a/crates/d2-little/packages/react-native/.gitignore
+++ b/crates/d2-little/packages/react-native/.gitignore
@@ -3,6 +3,7 @@
 ios/Frameworks/
 macos/Frameworks/
 android/src/main/jniLibs/
+windows/Frameworks/
 
 # react-native-builder-bob build output
 lib/
--- a/crates/d2-little/packages/react-native/package.json
+++ b/crates/d2-little/packages/react-native/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@actrium/supramark-d2-native-rn",
   "version": "0.1.0",
-  "description": "React Native native FFI wrapper for d2-little — D2 diagrams to SVG on iOS, macOS, and Android.",
+  "description": "React Native native FFI wrapper for d2-little — D2 diagrams to SVG on iOS, macOS, Android, and Windows.",
   "main": "lib/commonjs/index.js",
   "module": "lib/module/index.js",
   "types": "lib/typescript/index.d.ts",
@@ -21,6 +21,7 @@
     "ios",
     "macos",
     "android",
+    "windows",
     "scripts",
     "!android/.cxx",
     "!android/build",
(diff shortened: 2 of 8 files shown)


kookyleo (approved):
Reviewed and compiled on Windows MSVC. Produced supramark_d2_native.dll. ABI matches. MPL-2.0 headers agree with package.json.

Nits (non-blocking): same arm64 target typo and missing RNW autolink as #213/#214.

PR #206, opened 2026-08-17
fix(conformance): profile CommonMark-core cases with GFM autolink off

mybee:
## Summary

Resolve #202 and #203.

The CommonMark-core spec examples are normative CommonMark: bare URLs, bare emails, and the inner URL of an invalid bracketed autolink must stay literal. Supramark's shipped default enables the GFM bare-URL/email autolink extension (the product's GFM profile), so those cases diverged under the default config. The divergences were previously baselined (issue #144 stance) rather than fixed.

Profile CommonMark-core cases with `--no-gfm-autolink` (already supported by the CLI) so they are measured against CommonMark semantics:

- **commonmark source (0.31.2)** — entirely CommonMark-core, gated wholesale. Resolves `0602` / `0608` / `0611` / `0612` (#202).
- **cmark-gfm source (0.29.0.gfm.13)** — `spec.txt` CommonMark-core sections (those not suffixed `(extension)`) are gated, mirroring cmark-gfm's own `spec.txt` run which executes those sections without the autolink extension. The `(extension)` sections and the entire `extensions.txt` file keep the default config. Resolves `spec-0610` / `spec-0616` / `spec-0619` / `spec-0620` (#203).
- **micromark source** — unchanged (default config).

The product default (GFM autolink ON) is unchanged; `<Supramark>` still linkifies bare URLs/emails.

The shared `isCommonMarkCoreCase(testCase)` helper unifies the two sources' gating so the same rule applies wherever a case is normative CommonMark.

## Changes

- `scripts/run.mjs`: unified `isCommonMarkCoreCase(testCase)` gating + `parserProfile` label (`supramark-commonmark` for the commonmark source, `supramark-default` otherwise).
- `baselines/commonmark.json`: 0 failures, profile `supramark-commonmark`.
- `baselines/cmark-gfm.json`: 0 failures, notes updated (`spec-0610..0620` now RESOLVED).
- `lib/reports/issue-report.mjs`: report the actual profile instead of hardcoding "default parser configuration".

## Test plan

- [x] `cargo build -p supramark-markdown --bin supramark-markdown`; confirm the resolved cases emit a single `text` node under `--no-gfm-autolink`.
- [x] `node tests/markdown-conformance/scripts/run-visual.mjs commonmark` -> 652/652 semantic + visual, gate PASS.
- [x] `node tests/markdown-conformance/scripts/run-visual.mjs cmark-gfm` -> 702/702 semantic + visual, gate PASS.
- [x] `pnpm test:unit` (conformance harness) passes.

PR #223, opened 2026-08-27
fix(conformance): honest supramark-cmark-gfm profile label for cmark-gfm baseline

### LINKED ISSUE #220: follow-up: honest cmark-gfm parserProfile after #206
#206 landed the CommonMark-core GFM-autolink-off profile. Behaviour is correct (spec-0621 `Autolinks (extension)` stays on). Two documentation nits:

1. `tests/markdown-conformance/baselines/cmark-gfm.json` still claims `parserProfile: "supramark-default"` while 648/702 cases pass `--no-gfm-autolink`. Commonmark got an honest `supramark-commonmark` rename; cmark-gfm should not hide a split config under the default name (third label, or per-case profile in the summary).
2. The RESOLVED notes swap the four spec-* descriptions (0610/0616 are the invalid bracketed cases; 0619/0620 are the bare URL/email). Copied from the old baseline.

Related: #206, #202, #203.

mybee:
## Summary

Fixes #220.

- Introduce a third profile label `supramark-cmark-gfm`: `run.mjs` now emits it for the `cmark-gfm` source (CommonMark-core spec sections run `--no-gfm-autolink`, GFM extension sections run the default config), and `baselines/cmark-gfm.json` records it instead of hiding the split config under `supramark-default`. Mirrors the `supramark-commonmark` rename from #206.
- Correct the four swapped RESOLVED notes in the baseline `_summary`: 0610 (`<http://foo.bar/baz bim>`, space inside the brackets) and 0616 (`< http://foo.bar >`, leading space after the bracket) are the invalid **bracketed** autolink cases; 0619 (`http://example.com` bare URL) and 0620 (`foo@bar.example.com` bare email) are the bare-URL/email cases. The old notes had the two pairs swapped.

## Test plan

- [x] `node scripts/run-visual.mjs cmark-gfm` → semantic 702/702, visual 702/702, `gate[regression]: PASS`
- [x] Verified spec-case attribution against the upstream fixture (`cases.json` upstreamIds 608–621) before rewriting the notes


kookyleo (approved):
Verified locally: profile ternary in `scripts/run.mjs:88-94` matches `baselines/cmark-gfm.json:7`; `micromark.json` correctly stays on `supramark-default` (no split config), `commonmark.json` already renamed in #206. No stale hardcoded profile anywhere — `update-baseline.mjs` writes `summary.profile` dynamically, and `compareWithBaseline()` does a strict `parserProfile` match, so a mismatched label fails the gate loudly rather than relaxing it.

The four RESOLVED notes were checked against the fixture by upstreamId, not against the PR text: 0610 `<http://foo.bar/baz bim>` and 0616 `< http://foo.bar >` are the invalid bracketed-autolink cases; 0619/0620 are the bare URL/email cases. The rewrite is correct.

Ran `node scripts/run-visual.mjs cmark-gfm`: semantic 702/702, visual 702/702, `gate[regression]: PASS`, exit 0.

PR #224, opened 2026-08-27
fix(rn): inline image fallback, wrapper a11y label, failure reset + cache/diagram doc nits

### LINKED ISSUE #217: follow-up: RN image / cache / maxWidth nits after #200 #201 #211
Follow-ups from merged RN PRs. None of these blocked merge.

**Images (#200)** — `packages/renderers/rn/src/Supramark.tsx`

- Mixed-content / table / heading images still pass any URI to `Image` with no `onError`. Empty or relative src is a blank 20×20; the block `MarkdownImage` path already has `hasLoadableImageUrl` + placeholder. Reuse that on `renderInlineNode`.
- `TouchableOpacity` around block images (always for links, and for standalone when `onImagePress` is set) does not receive `accessibilityLabel`, so the inner `Image` label is easy to lose. Set the label (and decorative flags) on the wrapper; `accessible={false}` on the inner image.
- `failed` is not reset when `image.url` changes (gallery `key={index}`). Reset on URL change or key by URL.

**Render cache (#201)** — `packages/renderers/rn/src/renderCache.ts`

- `maxBytes: 0` is no byte cap, not “cache disabled”. `maxSize: 0` is the kill switch. Keep the JSDoc explicit so hosts do not use `maxBytes: 0` as a memory limit.
- No test that a cached AST root is `Object.isFrozen` after `getOrCreate`. Cheap assertion, not a behaviour change.

**Diagram maxWidth (#211)** — `packages/renderers/rn/src/DiagramNode.tsx`

- No test. `Dimensions` is mocked at 375 wide; `{ maxWidth: 200 }` vs unset on `SvgXml` width would lock the contract.
- Shared `SupramarkDiagramConfig.maxWidth` is ignored by `@supramark/web`.
- Units are RN dp; the `[0.6×cap, cap]` floor remains. Document that wrapping `<Supramark>` is not enough — hosts with a narrower bubble must pass the number.

Related: #200, #201, #211.

mybee:
## Summary

Fixes #217. Follow-ups from #200/#201/#211, all non-blocking nits.

**Images (`Supramark.tsx`)**
- Inline `renderInlineNode` image path now reuses `hasLoadableImageUrl`: an empty/relative/unloadable src renders the alt/title/`[image]` placeholder text inline instead of a blank 20×20 `Image`, matching the block `MarkdownImage` placeholder.
- The `TouchableOpacity` wrapper (link or `onImagePress`) now receives `accessibilityLabel` + decorative flags (`accessibilityElementsHidden` / `importantForAccessibility`); the inner image gets `accessible={false}` so the label is not lost.
- Per-URL failure state: `failedUrl` is compared against `image.url`, and `onError` records the URL — a gallery keyed by `index` resets the placeholder when the URL changes.

**Render cache (`renderCache.ts`)**
- `maxBytes` JSDoc now states explicitly: `0` means **no byte cap**, not "cache disabled"; the kill switch is `maxSize: 0` or `enabled: false`.

**Diagram maxWidth**
- `SupramarkDiagramConfig.maxWidth` JSDoc documents: RN-only (web ignores it), dp units, clamped to `[0.6×cap, cap]`, and wrapping `<Supramark>` narrower is not enough — hosts with a narrower bubble must pass the number.

## Test plan

- [x] `image.test.tsx` +7: inline relative/empty fallbacks; wrapper label for link/onImagePress/decorative cases; failure reset across URL change via `renderer.update`
- [x] `DiagramNode.test.tsx` +3: SvgXml width contract — 337.5 (=375×0.9) unset, 200 with `maxWidth`, 120 (=200×0.6) floor
- [x] `SupramarkCache.test.tsx`: parser mock captures the root; asserts `Object.isFrozen` after a cached render
- [x] Targeted runs of the three touched test files: 81/81 pass (3 pre-existing unrelated errors elsewhere in the package: missing wasm dist for parse-smoke, rn-selection mock interference)

### DIFF
--- a/packages/core/src/ast.ts
+++ b/packages/core/src/ast.ts
@@ -206,11 +206,14 @@ export interface SupramarkDiagramConfig {
   defaultTimeoutMs?: number;
 
   /**
-   * Host-provided cap on diagram display width (dp). Diagrams clamp their
-   * display width to min(windowWidth × 0.9, maxWidth). Hosts whose container
-   * is narrower than the window (e.g. a desktop chat bubble with a fixed
-   * maxWidth) pass their own cap so the diagram doesn't overflow the
-   * container. Unset means the screen-derived cap applies alone.
+   * Host-provided cap on diagram display width. React Native only —
+   * @supramark/web currently ignores this option and lays diagrams out with
+   * CSS. Units are RN dp, and the value is a cap, not a target: display width
+   * is clamped to [0.6 × cap, cap] where cap = min(windowWidth × 0.9,
+   * maxWidth), so small diagrams still get the 0.6× floor. Wrapping
+   * `<Supramark>` in a narrower container is NOT enough — the cap is derived
+   * from the window, so hosts with a bubble narrower than the window must pass
+   * the number explicitly or the diagram overflows the container.
    */
   maxWidth?: number;
 
--- a/packages/renderers/rn/__tests__/DiagramNode.test.tsx
+++ b/packages/renderers/rn/__tests__/DiagramNode.test.tsx
@@ -66,6 +66,17 @@ function successfulResult(): DiagramRenderResult {
   };
 }
 
+/** A wide diagram (intrinsic 1000×500) so display width is always cap-driven. */
+function wideResult(): DiagramRenderResult {
+  return {
+    id: 'test-wide',
+    engine: 'mermaid',
+    success: true,
+    format: 'svg',
+    payload: '<svg viewBox="0 0 1000 500"></svg>',
+  };
+}
+
 function failedResult(): DiagramRenderResult {
   return {
     id: 'test-error',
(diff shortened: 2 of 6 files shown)


kookyleo (approved):
All five claims verified against source, not the description.

The two commits do not conflict: the first stamps `isDecorative` straight onto the wrapper, which would have made a linked decorative image unreachable to screen readers; the second adds `&& !wrapped` and closes that. Final state is correct. New tests have real bite — reverting `Supramark.tsx` to main reds 6 of the 7 new cases. `tsc --noEmit`, `eslint --max-warnings 0` and `check-cjk` clean; RN suite 153 pass, the single failure being an unbuilt wasm dist unrelated to this PR. Scope is exactly the seven bullets of #217, nothing extra.

The `maxWidth` JSDoc matches `DiagramNode.tsx` (cap = min(screenWidth*0.9, maxWidth), chart = clamp to [0.6*cap, cap], web never reads it).

Two follow-ups filed separately rather than held against this PR: the `maxBytes: 0` contract this PR documents is not honored on the diagram-default path (pre-existing, from #201/#124), and the wrapper a11y work leaves an unnamed touchable plus an Android gap. Merging.

PR #225, opened 2026-08-27
fix(windows): aarch64 rustc triple, append CMAKE flags, prefer rustc import lib, MPL headers

### LINKED ISSUE #218: follow-up: Windows native port nits after #212–#215
Follow-ups from the Windows native ports now on `main` (#212–#215). Default `x86_64-pc-windows-msvc` builds produced DLLs on VS 2022 BuildTools. These are leftover holes, not regressions vs Android/iOS for the default arch.

1. `--arch arm64` becomes rustc target `arm64-pc-windows-msvc`, which does not exist. Use `aarch64-pc-windows-msvc` in `crates/{supramark-markdown,mermaid-little,d2-little,plantuml-little}/scripts/build-windows.sh`. graphviz-anywhere `build_helpers.rs` already uses that triple.
2. No RNW autolink project (`vcxproj`/`CMakeLists` + `react-native.config.js` `platforms.windows`). Same incomplete skeleton as in-tree graphviz Windows. `prepare-native.js` hardcodes `windows-x86_64`.
3. #212: `-DCMAKE_C_FLAGS="/utf-8"` **replaces** the cache variable. Use `"${CMAKE_C_FLAGS:-} /utf-8"` if extra C flags are ever passed in. PR CI's Graphviz Windows job still links a published release asset; `scripts/build-windows.sh` itself is not a required check.
4. #213: first `lib.exe /DEF:` against a `.def` that is never written (error swallowed). Prefer copying rustc’s `*.dll.lib`.
5. #214 Windows headers say Apache-2.0; mermaid-little RN package is MPL-2.0 / crate is MIT.

#216 (plantuml) is still open: its default `build-windows.sh` does not find `graphviz_api_static.lib`. Track that on the PR, not here.

Related: #212, #213, #214, #215.

mybee:
## Summary

Fixes #218 items **1, 3, 4, 5**. Item 2 (RNW autolink project skeleton + `prepare-native.js` `windows-x86_64` hardcode) is intentionally left out — it is the same incomplete skeleton as in-tree graphviz Windows and deserves its own PR; PR CI's Graphviz Windows job links a published release asset and does not gate on it.

1. `--arch arm64` now maps to the real rustc triple `aarch64-pc-windows-msvc` in `crates/{supramark-markdown,mermaid-little,d2-little}/scripts/build-windows.sh` (staged output dirs keep `windows-arm64` naming; same mapping as graphviz-anywhere `build_helpers.rs`).
3. `-DCMAKE_C_FLAGS` / `-DCMAKE_C_FLAGS` in graphviz-anywhere now **append** (`"${CMAKE_C_FLAGS:-} /utf-8"`) instead of replacing the cache variable.
4. `supramark-markdown` import-lib staging prefers copying rustc's own `supramark_markdown_native.dll.lib` (exactly matching the real exports); the lib.exe fallback keeps the vswhere discovery + minimal DEF, with the phantom `/DEF:` path that was never written removed.
5. mermaid-little Windows headers (`ReactPackageProvider.h/.cpp`, `SupramarkMermaidModule.h`) now carry `SPDX-License-Identifier: MPL-2.0`, matching the RN package.

## Test plan

- [x] `bash -n` on all four modified scripts
- [x] Import-lib path verified against rustc cdylib output naming (`<name>.dll.lib` next to the DLL)
- [x] No CI gate covers these scripts today (Graphviz Windows job links a published release asset) — the default `x86_64` path is unchanged in behavior

### DIFF
--- a/crates/d2-little/scripts/build-windows.sh
+++ b/crates/d2-little/scripts/build-windows.sh
@@ -7,7 +7,8 @@
 # then stage the DLL + import lib + C header into output/windows-x86_64/.
 #
 # Requires:
-#   - Rust target x86_64-pc-windows-msvc: rustup target add x86_64-pc-windows-msvc
+#   - Rust target (x86_64 default; aarch64 for --arch arm64):
+#       rustup target add x86_64-pc-windows-msvc aarch64-pc-windows-msvc
 #   - MSVC build tools (Visual Studio 2019+ or Build Tools)
 #   - Run on Windows (Git Bash / MSYS2) or CI windows-latest runner.
 #
@@ -32,12 +33,15 @@ while [[ $# -gt 0 ]]; do
 done
 
 case "$ARCH" in
-    x86_64|amd64) ARCH="x86_64" ;;
-    arm64|aarch64) ARCH="arm64" ;;
+    x86_64|amd64) ARCH="x86_64"; RUST_ARCH="x86_64" ;;
+    arm64|aarch64) ARCH="arm64"; RUST_ARCH="aarch64" ;;
     *) echo "Unsupported architecture: ${ARCH}"; exit 1 ;;
 esac
 
-RUST_TARGET="${ARCH}-pc-windows-msvc"
+# rustc has no arm64-pc-windows-msvc triple — the Windows-on-ARM target is
+# aarch64-pc-windows-msvc (same mapping as graphviz-anywhere build_helpers.rs).
+# Staged output dirs keep the windows-${ARCH} naming.
+RUST_TARGET="${RUST_ARCH}-pc-windows-msvc"
 BUILD_DIR="${BUILD_DIR:-${PROJECT_ROOT}/build/windows-${ARCH}}"
 INSTALL_DIR="${INSTALL_DIR:-${PROJECT_ROOT}/output/windows-${ARCH}}"
 
--- a/crates/graphviz-anywhere/scripts/build-windows.sh
+++ b/crates/graphviz-anywhere/scripts/build-windows.sh
@@ -136,13 +136,16 @@ mkdir -p "${BUILD_DIR}/graphviz"
 # MSYS2_ARG_CONV_EXCL: Git Bash converts the /utf-8 value as a Unix path into
 # C:/Program Files/Git/utf-8 (see CMakeCache pollution), so exclude exactly
 # these two flags from path conversion.
+# Append (not replace): -D on the command line overwrites any CMAKE_C_FLAGS
+# cache variable, so a caller exporting CMAKE_C_FLAGS keeps those flags by
+# folding them in ahead of /utf-8.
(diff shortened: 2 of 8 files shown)


kookyleo (approved):
Items 1, 3 and 4 verified against source rather than the description.

**Triple mapping** — all three `build-windows.sh` files add `RUST_ARCH` (`arm64` → `aarch64`) and use `RUST_TARGET` only for `rustup target add` / `cargo build --target` / artifact paths, while `BUILD_DIR` / `INSTALL_DIR` keep the old `windows-${ARCH}` naming. Searched the repo: no leftover `arm64-pc-windows-msvc` anywhere, and the staged directory convention is untouched, so `prepare-native.js` consumers still resolve. Matches graphviz-anywhere's `build_helpers.rs`.

**CMake flags** — both `cmake` invocations append for `CMAKE_C_FLAGS` **and** `CMAKE_CXX_FLAGS`; CXX was not missed. Quoting is right (the whole `-DVAR=value` is one bash argument; unset degrades to a harmless leading space), and `MSYS2_ARG_CONV_EXCL` already covers both names.

**Import lib** — `DLL_NAME` matches Cargo's dash→underscore lib name exactly, and `${DLL_SRC}.lib` is the right rustc/MSVC cdylib convention; the staged rename lands on the path `prepare-native.js` hardcodes. The second commit is a pure extraction — same diff moved into `generate_import_lib_fallback()`, with `local var; var=$(cmd)` split correctly so `set -e` still sees the exit code. `bash -n` clean on all four scripts; `shellcheck -x` clean on the largest one (the remaining SC2002 hints are pre-existing and outside the changed lines).

**Item 5 was fixed the wrong way, so I corrected it in `2e8b2cc2`.** The headers went from a wrong Apache-2.0 to an equally wrong MPL-2.0. `crates/mermaid-little/Cargo.toml:6` says `license = "MIT"` and `crates/mermaid-little/LICENSE` is the MIT text; the MPL-2.0 came from `packages/react-native/package.json:10`, which is itself a stale copy of the d2-little template (d2-little really is MPL-2.0, matching upstream D2 — mermaid is MIT). Issue #218 only pointed out the contradiction without ruling on which side wins. Now MIT in the three headers and in that package.json, so the next person copying from it copies the right one.

**One correction to the CI claim**, which matters for risk: the three crate-level `build-windows.sh` files genuinely have no workflow reference, but `graphviz-anywhere/scripts/build-windows.sh` *is* executed — `graphviz-release.yml` runs `./scripts/build-windows.sh --arch ${{ matrix.arch }}` on real `windows-latest` / `windows-11-arm` runners on tag pushes. It does not trigger on PRs, so nothing gates this change, but the CMake flag edit does get exercised at the next release rather than never.

Still unverified anywhere: aarch64 cross-compilation, the vswhere → lib.exe fallback chain, and CMake's cache behaviour under a real VS generator all need an actual Windows box.

PR #226, opened 2026-08-27
fix(markdown): defer inline-math scan until after adjacent-text reassembly

### LINKED ISSUE #219: follow-up: defer inline-math scan until after adjacent-text reassembly (#209)
#209 landed the slow path for `$…\{…\}…$` (#207). Isolated escaped-brace math parses. Two leftover holes, both pre-existing first-pass artifacts:

1. First-pass delimiter scan still fires on fragments. `$\{1\}$ and $\{2\}$` and `see $\{0\}$ and $x$` become a fake `math_inline` whose value is the interstitial `" and "` because `$ and $` is a length-aligned `Text` node. Defer `$` / `[^` scanning until after `rescan_adjacent_text_runs`. Same class: `:smile: $\{0\}$` (emoji replace poisons `RawValueMap`) and indented lazy continuation (`hello\n  $\{0\}$`).
2. Tests only cover a single math run. Add `$\{1\}$ and $\{2\}$` (xfail until (1) if needed) and a negative `\$not math$`. `find(math_inline)` will not notice an extra bogus math node.

Related: #209, #207.

mybee:
## Summary

Fixes #219. Follow-up to #209/#207.

The first-pass `$` / `[^` scan ran per TextScanner fragment, so a length-aligned trailing fragment paired its `$` with the opening `$` of a *different* span — `$\\{1\\}$ and $\\{2\\}$` produced a fake `math_inline` whose value was the interstitial `" and "`. Emoji expansion had the same defect: replacing `:smile:` before the rescan desynchronized the value↔raw byte alignment and collapsed the whole run to plain text.

- Text/TextSpecial `to_ast_v2` now emits plain `Text` via a new `map_text_fragment` — no scanning, no emoji expansion on the first pass.
- Scanning happens exactly once in `flush_text_run`, after `rescan_adjacent_text_runs` has reassembled each contiguous run; emoji shortcodes expand on the emitted text slices only.
- `RawValueMap` models every decoding a whole run can now contain, so it stops bailing (which sinks the whole run to plain text):
  - whitespace cmark strips from the value but that still occupies the source span (continuation-line indentation) — `hello\\n  $\\{0\\}$`
  - HTML entities / numeric char refs via `decode_entity_at_start`, mirroring the parser's entity grammar exactly (hex `{1,6}` / decimal `{1,7}`, invalid charcode → U+FFFD, case-sensitive named table) — `a &amp; b $x$`, `&copy; $\\{0\\}$`, `&#0; costs $x$`
  - CR / CRLF line endings (raw-only `\\r`, incl. the run's trailing line ending) — `a $x$\\r\\nb $y$`
- Math content keeps backslash escapes raw (TeX needs `\\{` verbatim) but decodes entities, so `$a &lt; b$` carries `a < b` to the TeX engine; the delimiter scan itself still runs on the raw slice.

## Test plan

- [x] Regression tests in `crates/supramark-markdown/tests/public_api.rs` (deduplicated via `paragraph_children` / `math_values` helpers): `$\\{1\\}$ and $\\{2\\}$`, `see $\\{0\\}$ and $x$` (values + no stray `$` in text), `:smile: $\\{0\\}$` (emoji then math), `hello\\n  $\\{0\\}$` (lazy continuation), `\\$not math here` (negative), entity matrix (`&amp;` / `&copy;` / `&#65;` / literal `&unknown;`), CRLF matrix (multi-line + CRLF-with-indentation), invalid numeric refs (`&#0;` / `&#xD800;` / `&#x110000;` and beyond-grammar `&#x0012345;` staying literal), entities inside math content (`$a &lt; b$` → `a < b`, escapes stay raw)
- [x] Both follow-up regressions verified against the parent commit's binary before fixing (old code parses the math; the first commit of this branch lost it)
- [x] `cargo test -p supramark-markdown`: 140/140 pass (86 lib + 54 public_api)
- [x] Conformance gates (binary rebuilt first): `run-visual.mjs cmark-gfm` 702/702 PASS, `commonmark` PASS, `micromark` PASS (no regressions vs baseline)
- [x] `bun run test:core` 101/101; math/emoji/footnote feature tests green

### DIFF
--- a/crates/supramark-markdown/src/common/utils.rs
+++ b/crates/supramark-markdown/src/common/utils.rs
@@ -13,6 +13,8 @@ static DIGITAL_ENTITY_TEST_RE: Lazy<Regex> =
     Lazy::new(|| Regex::new(r#"(?i)^&#(x[a-f0-9]{1,8}|[0-9]{1,8});$"#).unwrap());
 static UNESCAPE_ALL_RE: Lazy<Regex> =
     Lazy::new(|| Regex::new(&format!("{UNESCAPE_MD_RE}|{ENTITY_RE}")).unwrap());
+static ANCHORED_ENTITY_RE: Lazy<Regex> =
+    Lazy::new(|| Regex::new(&format!("^{ENTITY_RE}")).unwrap());
 
 #[allow(clippy::manual_range_contains)]
 /// Return true if a `code` you got from `&#xHHHH;` entity is a valid charcode.
@@ -97,6 +99,16 @@ fn replace_entity_pattern(str: &str) -> Option<String> {
     } else {
         None
     }
+}
+
+/// Decode a valid HTML entity / numeric character reference that starts at
+/// the beginning of `raw` (`&amp;`, `&#65;`, `&#x41;`), returning the decoded
+/// replacement and the raw byte length consumed. Returns `None` when `raw`
+/// does not start with a decodable entity.
+pub(crate) fn decode_entity_at_start(raw: &str) -> Option<(String, usize)> {
+    let matched = ANCHORED_ENTITY_RE.find(raw)?;
+    let replacement = replace_entity_pattern(matched.as_str())?;
+    Some((replacement, matched.as_str().len()))
 }
 
 /// Unescape both entities (`&quot; -> "`) and backslash escapes (`\" -> "`).
--- a/crates/supramark-markdown/src/common/utils.rs
+++ b/crates/supramark-markdown/src/common/utils.rs
@@ -141,29 +141,6 @@ pub(crate) fn decode_entity_at_start(raw: &str) -> Option<(String, usize)> {
     Some((decoded.to_string(), matched.as_str().len()))
 }
 
-/// Decode every entity in `raw` that the inline entity scanner would have
-/// consumed (one level, matching the parser). Used for inline-math content
-/// carved from the raw source: backslash escapes must stay raw for TeX, but
-/// an entity the parser already decoded in surrounding text should reach the
-/// math engine decoded too (`$&lt;$` is the math `<`, not the four bytes
-/// `&lt;`).
-pub(crate) fn decode_entities(raw: &str) -> String {
-    let mut out = String::with_capacity(raw.len());
-    let mut rest = raw;
-    while let Some(amp) = rest.find('&') {
-        if let Some((decoded, len)) = decode_entity_at_start(&rest[amp..]) {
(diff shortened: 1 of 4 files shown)


kookyleo (approved):
The core fix is bigger than the description claims, and I verified it by building the PR and an origin/main binary side by side:

```
input                     PR              main
$\{1\}$ and $\{2\}$      two correct spans   ONE fake span, value " and "
```

And it is not limited to paragraphs — inline math inside headings, table cells, link text, emphasis/strong/strikethrough, block quotes, list items and footnote definition bodies is **entirely broken on main** (collapses to plain text) and works here. Differential fuzzing (200k cases) produced 1175 differences, every one of them a recovery: 900 math, 169 position, 5 leftmost-match `[^…]`, and 99 emoji no longer expanding inside a split footnote label. Not one capability lost. 2505 real fixture documents diff byte-identical at AST level. Performance improves rather than regresses, because main was scanning twice (per fragment in `skip_text`, then per run in `flush_text_run`) and this removes the first pass; `inline_extension_scan_needed` short-circuits before any `RawValueMap` is built. No UTF-8 slicing panic is reachable: `&raw[ri..]` is only evaluated at an ASCII `&`, and every `to_value()` call site is an ASCII boundary — 550k fuzz cases plus the real corpus, zero panics.

**One change I removed before merging (`2f52892a`).** Commit 3 decoded HTML entities in the carved math span. That let a decoded value walk around the scanner's own rule: the delimiter scan runs on the raw slice and rejects a newline inside `$…$`, so `$a\nb$` yields no math — but `$a&#10;b$` produced `math_inline` with value `a\nb`, `&#13;` put a bare `\r` in the AST, and `&dollar;` put an unescaped `$` into a value headed for the TeX engine. It also contradicted commit 2 of this same branch, which said the value keeps `&amp;` "matching micromark-extension-math" — and commit 2 was right: upstream `math-text.js` swallows everything between the delimiters as one `mathTextData` token with no sub-tokenizing, so entities are not decoded, exactly like a code span. MathBlock in this crate does not decode either, so inline and block had ended up with opposite behaviour. Math content is now taken verbatim; the regression test is inverted and extended with the three smuggling cases. 140/140 pass, clippy unchanged at 16 pre-existing warnings, and all three conformance gates PASS in CI.

The remaining follow-up — the entity grammar being hand-copied into `RawValueMap` rather than shared with `entity.rs`, where a decimal-length drift silently drops math with the whole test suite still green — is tracked in #234 rather than held against this PR.

PR #229, opened 2026-08-28
test(markdown): add cmark regression conformance cases

alpacassarg:
- Add 27 cmark regression cases in the unified JSON format
- Preserve the CR+CR+LF line-ending regression case
- Record the pinned upstream commit and fixture checksums
- Add source configuration, importer, validation, and baseline support
- Support `cmark-regression` individually and through `all` in the workflow
- Update English and Chinese documentation


kookyleo (approved):
Solid work on the plumbing, verified rather than taken on trust: the importer reproduces the fixture byte-for-byte from the pinned upstream commit (`git diff --exit-code` clean after a live re-import), `7042d997` really exists on cmark's master, and the recorded sha256 matches `test/regression.txt` bit for bit. The CR+CR+LF case is correct — `cases.json` holds `line1\r\r\nline2\n` (not `\r\n\n`), the `inputContains` integrity check really asserts it, and the `spec-examples.mjs` regex change (`.*` → `[^\n]*`) is a genuine bug fix, since JS `.` excludes `\r` and the old form silently dropped those bytes. No `|| true` anywhere, exit codes propagate, REUSE licence files present, and the source is picked up automatically by `resolve-sources.mjs` so it really does run on PR CI.

One thing had to be fixed before merging. The baseline declared 13 semantic / 10 visual failures while the suite reports 7 / 6, so six passing cases (0003, 0006, 0008, 0019, 0025, 0026) were pre-approved as known failures — and the regression gate only reacts to newly added failures, never to resolved ones, so those six would have been unguarded from day one. They are not marginal either: raw HTML blocks, emphasis delimiter runs, list-item thematic breaks, CDATA and mixed-case HTML blocks. The number was already printed in this PR's own CI log as `6 resolved`. Regenerated with `run-visual.mjs` + `update-baseline.mjs`; CI now reports `0 resolved, 7 still failing`.

The 7 real failures and two doc nits are tracked in #233 rather than held against this PR.

Worth knowing separately: none of the four conformance jobs is in the `main` ruleset's required checks, so this gate is advisory today — which is exactly why the baseline needed to be honest.

PR #216, opened 2026-08-20
feat(windows): plantuml diagram engine React Native Windows port

boomsi:
## Motivation

`@actrium/supramark-plantuml-native-rn` currently has native bindings for iOS / Android / macOS only. On Windows, the `createReactNativeDiagramEngine` lookup fails to find a Windows adapter and falls back to the wasm loader — which has not been verified to work in Hermes.

## Change

Add Windows native bindings for the plantuml engine, following the same pattern as the iOS / Android / macOS bindings:

**New C++/WinRT module** — `crates/plantuml-little/packages/react-native/windows/SupramarkPlantumlNative/`:
- `ReactPackageProvider.cpp` / `.h`: registers the module with the RNW runtime
- `SupramarkPlantumlModule.h`: defines the FFI signature, `dllimport` function declarations, and the `ReactPackage` template

**JS layer**:
- `package.json`: `files` adds `windows`, `peerDependenciesMeta` adds `react-native-windows`, adds a `react-native-windows` platform extension
- `src/index.ts`: `Platform.select` adds a `windows` branch
- `scripts/prepare-native.js`: Windows binary staging (copy `.dll` / `.h` / `.lib` to `windows/Frameworks/`)

**`gitignore`**: `windows/Frameworks/` added (consistent with existing `ios/Frameworks/`, `macos/Frameworks/`).

**Build script** — `crates/plantuml-little/scripts/build-windows.sh` (122 lines): cross-compiles the Rust crate to a Windows `.dll` via the C ABI. Note: plantuml's native implementation depends on a Java runtime, so this script also fetches/configures the JRE.

## Risk

- Windows-only — no impact on iOS / Android / macOS
- 8 files, 299 lines
- FFI signature must match the Rust crate's `extern "C"` exports
- JRE availability on the build machine — the script must handle a missing/incompatible JRE gracefully
- `prepare-native.js` Windows branch handles `x86_64` and `arm64` architectures

## Test plan

- [ ] Run `./build-windows.sh` in `crates/plantuml-little` — expect JRE fetched and `output/windows-x86_64/bin/supramark_plantuml_native.dll` produced
- [ ] Run `npm run prepare-native` — expect `windows/Frameworks/{bin/*.dll, include/*.h, lib/*.lib}` populated
- [ ] In a host RNW app, side-effect `import '@actrium/supramark-plantuml-native-rn'` at entry top — expect `createReactNativeDiagramEngine()` resolves to the native adapter
[...]

PR #236, opened 2026-09-07
ci(conformance): add a stable Conformance gate check

kookyleo:
## Why

None of the conformance jobs is in the `main` ruleset's required checks, so the gate is advisory today — a red conformance run does not stop a merge. #229 is the concrete case: it landed a baseline declaring 13 semantic / 10 visual failures while the suite actually reports 7 / 6, pre-approving six passing cases (raw HTML blocks, emphasis delimiter runs, list-item thematic breaks, CDATA / mixed-case HTML blocks) as known failures. The regression gate only reacts to newly *added* failures, so those six would have been unguarded from the day they landed, and nothing in CI would have said so.

## Why not just require the four existing checks

Two ways that breaks:

1. The matrix job names come from the data sources (`${{ matrix.source }} semantic and visual conformance`). #229 has just added a fourth source; a fifth, or a rename, leaves a required name that never reports.
2. The `pull_request` trigger carried the same `paths` filter as `push`, so the workflow does not run for changes outside six paths. A required check that never reports leaves the PR stuck on "Expected — waiting for status" with no way to clear it. #205 and #216, both open right now, touch none of those paths.

## What this does

- Adds one `Conformance gate` job that `needs` the whole matrix. GitHub aggregates a matrix dependency to `success` only when every leg succeeded, so the gate is red if any source fails, and `if: always()` keeps it reporting when a leg fails or the matrix is skipped.
- Drops the `paths` filter from `pull_request` only, so the gate always reports on a PR. The `push` trigger keeps its filter.

After this merges, `Conformance gate` is the single name to add to the ruleset's required checks.

## Cost

Every PR now runs the four conformance sources instead of only path-matching ones. They run in parallel and the suite is the same one that already runs on the majority of PRs today.

### DIFF
--- a/.github/workflows/markdown-conformance.yml
+++ b/.github/workflows/markdown-conformance.yml
@@ -46,14 +46,12 @@ on:
   # The issue-filing step already carries a `github.event_name !=
   # 'pull_request'` guard (dead until now), so PR runs report through the
   # check and the run artifacts, and never open or update an issue.
+  # No `paths` filter here, unlike the push trigger. `Conformance gate` below
+  # is a required check on main, and a required check that never reports
+  # leaves the pull request waiting forever. A path filter would do exactly
+  # that for any pull request outside the six paths above, so the workflow
+  # runs on every pull request and the gate always reports.
   pull_request:
-    paths:
-      - ".github/workflows/markdown-conformance.yml"
-      - "tests/cases/**"
-      - "tests/markdown-conformance/**"
-      - "crates/supramark-markdown/**"
-      - "packages/core/**"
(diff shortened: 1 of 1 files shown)


g65537 (approved):
Approving as the second maintainer account so this governance change can land; the review itself was done under kookyleo.

The aggregate-job approach is the right one: matrix job names track the data source list and would break the ruleset on the next source change, and the pull_request path filter would have left any PR outside those six paths waiting forever on a check that never reports. `Conformance gate` reported SUCCESS on this run alongside all four sources.

PR #237, opened 2026-09-22
fix(graphviz-rust): download release assets by default and link system libs by SONAME

kookyleo:
## Problem

`cargo install markon markond` (and any other downstream binary that pulls in `graphviz-anywhere` from crates.io) fails in `build.rs`:

```
graphviz-anywhere: unable to locate graphviz_api native library.
```

The published crate ships an empty `prebuilt/` and there is no repo `output/` tree, so the only viable path is the GitHub release download, which was opt-in via `GRAPHVIZ_ANYWHERE_ALLOW_DOWNLOAD=1` (introduced in c48a49b2 when no release feed existed yet).

After enabling the download, linking still failed on hosts without the -dev packages:

```
mold: fatal: library not found: expat
```

## Changes

- `build.rs`: the release download is now the default last-resort fallback. `GRAPHVIZ_ANYWHERE_NO_DOWNLOAD=1` still disables it; `GRAPHVIZ_ANYWHERE_ALLOW_DOWNLOAD` is accepted as a no-op. `publish-cargo` runs after `release`, so the tag for the crate version always has assets.
- `build.rs`: on Linux, link `libstdc++.so.6`, `libexpat.so.1` and `libz.so.1` by SONAME via `+verbatim`, so only runtime packages are needed. MSRV bumped to 1.67 (verbatim stabilization; build.rs already used let-else from 1.65).
- Docs: release links pointed at the defunct `Actrium/graphviz-anywhere` repo.
- Release 0.2.6.

## Verification

- `cargo package`, extracted the `.crate`, built an external consumer with no env vars on a host without `libexpat1-dev`: downloaded v0.2.5 assets, linked, rendered DOT → SVG; `ldd` shows `libexpat.so.1` / `libz.so.1` / `libstdc++.so.6`.
- `GRAPHVIZ_ANYWHERE_NO_DOWNLOAD=1` still fails with the descriptive panic.
- `cargo test` (65 tests), `cargo clippy --all-targets`, `cargo fmt --check` clean.


g65537 (approved):
Reviewed the three commits and the full diff; the description matches the code.

Verified:
- `graphviz-release.yml`: `publish-cargo` needs `release`, which needs every build job, so the `v<version>` tag always carries the assets before the crate is published; asset names and the `v{CARGO_PKG_VERSION}` URL match `target_triple_to_asset_name` and the release file list (v0.2.5 has all 14 assets).
- `build.rs`: only the opt-in gate is removed; `GRAPHVIZ_ANYWHERE_NO_DOWNLOAD=1` still short-circuits before any network access, the env override / prebuilt / repo `output/` paths run first, wasm32 returns early, and Android/iOS/macOS/Windows link lines are unchanged (the SONAME block is gated on `target_os == "linux"`, so Android keeps its empty list).
- `+verbatim` was stabilized in Rust 1.67.0 (rust-lang/rust#104360) and build.rs already used let-else (1.65), so `rust-version = "1.67"` is accurate.
- Local check on a host that has `libexpat.so.1` / `libstdc++.so.6` but no `libexpat.so` / `libstdc++.so` dev symlinks: the base build.rs fails at link time, the PR build.rs downloads the v0.2.5 asset, links, renders SVG, and `ldd` resolves the three SONAMEs; `GRAPHVIZ_ANYWHERE_NO_DOWNLOAD=1` still gives the descriptive panic; `cargo test -p graphviz-anywhere` passes (65 tests); `cargo metadata --locked` is clean.
- rustls 0.23.45 is the patched version for RUSTSEC-2026-0285 / GHSA-2mjx-qc3c-rqvc; the lockfile change is limited to rustls + rustls-webpki.
- Version, CHANGELOG entry and links, and doc release URLs are consistent; all checks are green.

Non-blocking nits, fine to leave for a follow-up:
- `crates/graphviz-anywhere/examples/rust/Cargo.toml` still says `rust-version = "1.64"` while depending on a 1.67 crate.
- `GRAPHVIZ_ANYWHERE_ALLOW_DOWNLOAD=1` in `ci.yml` (release runtime smoke) and in `publish-cargo` is now a no-op and could be dropped.
- Until the v0.2.6 tag exists, an in-repo build without a local `output/` tree will attempt the v0.2.6 download and fall through to the same panic as before, just with one extra curl attempt.

PR #238, opened 2026-09-22
fix(font-metrics): measure missing CJK glyphs at 1em under Notdef policy

kookyleo:
## Problem

kookyleo/markon#102: CJK message labels in a PlantUML sequence diagram render with their glyphs piled on top of each other.

```plantuml
@startuml
actor User
participant Client
User -> Client: 输入账号密码
Client --> User: 登录成功
@enduml
```

The embedded DejaVu Latin subset has no CJK glyphs, so under `MissingGlyphFallback::Notdef` (the policy plantuml-little uses) every ideograph measured as `.notdef`, about 0.6em. The six-character label above got `textLength="46.8076"` at 13px (7.8px per char) instead of 78. Browsers draw CJK at 1em and, with `lengthAdjust="spacing"`, squeeze the glyphs back into the short `textLength`, so they overlap; participants were also spaced too tightly for the labels.

Java PlantUML measures on AWT's composite `SansSerif` font, which falls through to a CJK slot where these chars are 1em.

## Change

`crates/font-metrics/src/ttf_parser.rs`: under the `Notdef` policy, a missing char with East Asian Width W/F (Hangul, Kana, Han incl. extensions, Yi, fullwidth forms — `is_east_asian_wide`) measures 1em. The `Space` policy used by mermaid-little is untouched; d2 does not use this crate.

## Verification

- New `cjk_message_label_is_measured_full_width` in `plantuml-little/tests/port_sequence_render.rs`: labels measure 78 / 52px and the lifeline gap fits them. Fails before the fix (46.8076).
- Two font-metrics unit tests (Notdef → 1em for Han/Kana/Hangul/fullwidth; Space policy unchanged).
- `cargo test -p plantuml-little`: 4416 passed, 0 failed; the Java reference-SVG suite needed no golden changes (no fixture contains CJK). font-metrics 9/9, mermaid-little green.
- Rendered with resvg + a CJK fallback font: before reproduces the issue screenshot; after the labels are clean, viewBox width 458 → 490.

To reach markon this needs a `supramark-font-metrics` 0.1.x patch release (plantuml-little depends on `^0.1`).


g65537 (approved):
Reviewed the diff against the base code and verified the claims locally.

- Root cause holds: `char_advance` only reaches the new branch when the char is absent from the face's cmap, so callers that supply a real CJK font are unaffected; under `Notdef` the embedded DejaVu Latin subset measured ideographs at the `.notdef` advance (~0.6em), and Java's composite `SansSerif` measures them on a CJK slot at 1em, which is also what browsers draw.
- Blast radius checked: only plantuml-little constructs `TtfParserMetrics` with the default `Notdef` policy. mermaid-little opts into `Space` (its CJK fixtures are unchanged, `space_fallback_keeps_east_asian_wide_at_space_width` locks that), d2-little only uses the `Metrics` trait via its own emulation/host impls, and the host/FFI callback paths never enter `char_advance`. No plantuml reference fixture contains CJK, so the reference-SVG suite is unaffected (confirmed green in CI).
- Ran `cargo test -p supramark-font-metrics` (9/9) and `port_sequence_render` (8/8) locally; with the base `ttf_parser.rs` the new test fails with `46.8076` as described, so it really guards the fix.
- `is_east_asian_wide` cross-checked against Unicode 16 EastAsianWidth: every listed range is W/F apart from U+3248..U+324F (circled numbers on black square, EAW=A) which ride along in the `3041..33FF` span. Emoji are deliberately excluded, keeping `missing_glyph_notdef_fallback` valid.

Non-blocking nits, fine to leave for a follow-up:
- U+3248..U+324F are EAW=A rather than W; harmless but the doc comment says W/F only.
- A few W blocks CJK fonts do cover are not listed: Enclosed Ideographic Supplement (U+1F200..U+1F265), Kana Supplement / Extended-A/B (U+1B000..U+1B16F). Rare in diagram labels.
- The description says d2-little does not use this crate; it does depend on `supramark-font-metrics` for the `Metrics` trait, it just never routes through `TtfParserMetrics`, so the conclusion stands.

CI is fully green. Approving.

PR #239, opened 2026-09-22
fix(mermaid-little): size labels for every line the browser paints

kookyleo:
Mermaid labels are measured as a single line, but a browser breaks a label at
every `<br/>`. A node box was therefore one line tall whatever the label, and
the second line painted below the box (markon#97); multi-line edge labels were
under-reserved in layout and crowded their neighbours.

Label height is now one line-height per painted line, in layout and in the
emitted `foreignObject`, for flowchart node labels (all shapes), flowchart edge
labels, state-diagram notes and state-diagram edge labels. Widths are left
exactly as they were: switching them to the widest line exposed the shim's CJK
under-measurement (CJK glyphs measure as a space), so that belongs in its own
change.

Line counting lives in one helper, `label_metrics::split_label_lines`, and both
label spellings are counted on the markup the renderer will actually measure
(`string_label_to_html` / `markdown_label_to_html`, then `replace_fa_icons`), so
layout and render agree by construction -- an FA icon alone on a line paints no
line box, and both sides know it. A markdown label's width comes from plain text
that is never re-parsed as markup, so a literal `<` in ``A["`i<n`"]`` stays text;
FA icon tokens are stripped per line, before the lines are joined, so an icon
name ends at its break instead of eating the next line. For genuine markup
input, where a tag starts is decided in one place (`label_metrics::tag_len`: `<`
plus a letter, `</` plus a letter, or `<!`, and never spanning a `<` or a line
break), which also fixes `x < y > z`, mis-measured on `main`.

The splitter's rules were measured in headless Chromium on the real label div
(`display: table-cell; white-space: nowrap; line-height: 1.5`):

- `<br>` in any spelling (`<br/>`, `<br />`, `<BR>`, `<br class="x">`) and a
  source `\n` (upstream rewrites it to `<br/>`) end a line.
- A `<p>` boundary ends a block, so markdown paragraphs (`<p>a</p><p>b</p>`,
  and `one<p>two`) are two lines.
- A break at the end of a block opens no line box: `a<br/>` is one line,
  `a<br/><br/>` two, `a<br/><br/>b` three, `<br/>a` two, `<br/>` one.
- A block with no text and no break (`<p></p>`) paints nothing.

Markdown labels keep their line breaks through measurement, so the layout box
grows with the label instead of only the label growing. State notes measure
after the `\n` -> `<br/>` rewrite so both note syntaxes agree, and the note box
grows to match.

Verification: every `.mmd` under `tests/fixtures` + `tests/ext_fixtures` (1396)
rendered with `main` and with this branch. 84 differ -- 39 cypress/flowchart,
30 demos/flowchart, 13 cypress/state, 2 demos/state -- and every one of them has
a `<br>`, a source newline, or a multi-line note. Loading each rendered SVG in
Chromium and comparing every label's declared `foreignObject` height against its
painted height: `main` mismatches on 397 fixtures, this branch on 313 -- the same
84 fixtures fixed, none regressed. The rest are pre-existing and untouched
(other diagram types still emit the jsdom 16.296875 line height; flowchart
84/118/166/217 are icon font-size and `max-width: 200px` wrapping).

Regression tests in `tests/diagram_bounds.rs` cover the node box containing its
label, multi-line edge labels, markdown labels, the trailing-`<br/>` rule, state
notes (both syntaxes, label and box) and state edge labels; unit tests in
`label_metrics` pin the browser line-count table. `cargo test -p mermaid-little`
1044 pass, clippy `-D warnings` clean, fmt clean, reference sweep unchanged at
704/1323.

Known and not addressed here: a `<br/>` in a subgraph title still renders as
escaped text (same on `main`); multi-line boxes stay as wide as the concatenated
lines; and a `classDef font-size` override still offsets the label group by the
default-size half height (`main` mis-centres both axes the same way -- correcting
it changes geometry for every font-size-override node).


g65537 (changes requested):
Thanks for the thorough write-up. I built the CLI at `origin/main` and at `524f15f` and rendered every `.mmd` under `tests/fixtures` + `tests/ext_fixtures` with both (1396 fixtures): 1322 are byte-identical, the 74 that differ are 69 flowchart fixtures whose labels contain `<br>`/newlines and 5 state fixtures with `<br/>` notes. The issue's diagram, all four `<br>` spellings, blank lines, string labels with literal newlines, and multi-line edge labels (incl. markdown edge labels) all come out right: `foreignObject` = 24 x lines, the node box grows, the label fits for rect / diamond / hexagon / stadium / cylinder / circle / flag, and dagre reserves the extra edge-label height. CI is green. Three things need another pass before this can go in:

**1. Markdown node labels are not fixed in layout (description says they are).**
`measure_vertex_box` runs markdown labels through `strip_markdown_for_measure` -> `strip_markdown_paragraph_for_measure`, which drops every `\n` and skips every `<...>` tag *before* `split_html_into_lines` gets to count lines. So for `` A["`line1<br/>line2`"] `` or a backtick label with a real newline the layout still sizes one line, while `shape_label_block_inner` now emits a 48px `foreignObject`. Result: box 46.3px, label 48px, still overflowing. Visible in cypress flowchart 174/45 (`` b("`The dog in **the** hog.(1)\nNL`") ``): the FO goes 24 -> 48 but the node rect is unchanged. The description lists "newlines in markdown labels" as part of the bug and says each extra line adds one HTML line in layout, so this case should either be covered (keep `\n` / `<br>` as line separators in the markdown stripping helper, and join paragraphs with a break) or explicitly called out as not addressed.

**2. A trailing `<br/>` is counted as an extra line.**
`split_html_into_lines`, `plain_text_lines` and `parse_html_text_lines` all push an empty last line when the label ends with `<br/>`, but a line break at the end of a block does not create a line box in a browser (`<p>a<br/></p>` paints one line; only `<br/><br/>` gives an empty line). With this PR `A["trailing<br/>"]` gets a 48px FO in a 70px box, and cypress state 13/55 (`note left of State2 : This is the note +,-<br/>`) get a 48px FO whose single painted line now sits 12px above the note centre. Dropping the final empty line when the markup ends with a `<br>` (while keeping `a<br/><br/>b` at three lines) would match the browser.

**3. State-diagram notes change too, and only halfway.**
`render/shapes/note.rs` measures the note through `measure_html_markup_label`, so this PR changes cypress state 13/15/55/57 and demos state 10, which the description does not mention (the "704/1323 unchanged" number only holds because those fixtures were already non-exact). The change is also inconsistent inside one diagram: a note written with a literal `<br/>` now gets `lines x 24` (cypress 15: 120px), but a multi-line `note ... end note` block is still measured as one line because its `\n` is only turned into `<br/>` by `normalize_note_html` *after* measurement, and the note rect in `layout/state.rs` (`html_label_line_height(DEFAULT_FONT_SIZE) + 2 * NOTE_PAD`) never grows either way, so the taller label is just re-centred on a one-line box. Please either keep this change scoped to the flowchart paths or make the note path consistent (measure after `\n` -> `<br/>` and grow the note box), and mention the state fixtures in the description.

Nits (non-blocking):
- `is_br_tag` / `plain_text_lines` / `split_html_into_lines` don't recognise a `<br>` with attributes (`<br class="x">`); probably fine, just noting.
- There are now three near-identical "split HTML at `<br>`" helpers across `layout/flowchart.rs`, `layout/label_metrics.rs` and `render/foreign_object.rs`; sharing one would keep the trailing-`<br>` rule from drifting.
- Not this PR: a `<br/>` inside a subgraph title is rendered as the literal escaped text `sub&lt;br/&gt;title` (same on main).

g65537 (changes requested):
Re-reviewed at `4c17347` (base `4d4d7ad`). The three points from the last round are fixed and I could reproduce the description's numbers, but the markdown change introduces one new regression that has to go before this merges.

**What I verified**

- Built `main` and this head, rendered all 1395 `.mmd` fixtures with both: 84 differ (39 cypress/flowchart, 30 demos/flowchart, 13 cypress/state, 2 demos/state), each one has a `<br>`, a newline in a label or a multi-line note. Loading every render in headless Chromium and comparing each `foreignObject`'s declared height with the painted div: files with a mismatch go 408 -> 324 (my harness counts slightly differently from yours, but it is the same 84 fixtures), no fixture gets worse, and no fixture with a changed render still mismatches.
- The browser rule table in `split_label_lines` matches Chromium on the label div (`a<br/>`=1, `a<br/><br/>`=2, `a<br/><br/>b`=3, `<br/>a`=2, `<br/>`=1, `a<br> <br>b`=3, `<p>a</p><p>b</p>`=2, `<p>a<br/></p><p>b</p>`=2, `<p></p><p>a</p>`=1, `<br class="x">`/`<BR>`/`<br />` all break, `<brand>` does not).
- Blocker 1 (markdown labels): `` A["`md<br/>br`"] ``, a backtick label with a real newline (cypress 174/45) and a two-paragraph label all now get a 48px FO **and** a 70.3px box; cypress/flowchart/46 (4 lines) grows to 118px. Fixed in layout, not only in render.
- Blocker 2 (trailing `<br/>`): `trailing<br/>` = 24/46.3, `a<br/><br/>` = 48, `<br/>lead` = 48, `a<br/><br/>b` = 72, `edge<br/>` = 24, `note … : single<br/>` = 24 in a 54px note. Matches the browser.
- Blocker 3 (state notes): `note … end note` (2 lines) and `: x<br/>y<br/>z` both measure per line (48 / 72) and the note box grows to label + 30 (78 / 102); a leading `<br/>` and a blank line inside a note block also agree with Chromium. The state edge-label commit is consistent between `layout/state.rs` and `svg_state.rs` (`one<br/>two` = 48, `one <br>two<br>three` = 72, trailing `<br/>` = 24, literal `\n` stays 1) and the `labelBkg` covers all lines; that change is justified — before it the second line painted below the background.
- Other callers of `measure_html_label` / `measure_html_markup_label` (class, requirement, sequence KaTeX, cluster titles) render byte-identically across the sweep.
- `cargo test -p mermaid-little`: 1040 pass; `cargo clippy --all-targets -D warnings` and `cargo fmt --check` clean; CI green.

**Blocking: markdown labels with a bare `<` lose width in layout**

`strip_markdown_for_measure` now wraps each paragraph in `<p>…</p>`, and `concat_label_text_for_width` treats *any* `<` as a tag start and skips to the next `>`. On `main` a bare `<` in a markdown label had no `>` to pair with and stayed literal; now the closing `</p>` supplies one, so everything from the `<` to `</p>` is dropped from the width measurement:

```
flowchart TB
    A["`a < b`"]
    C["`a <`"]
    D["`< b`"]
    F["`a < b < c`"]
    H["a < b"]
```

| node | rect width on `main` | rect width at `4c17347` | FO width |
|---|---|---|---|
| A ``"`a < b`"`` | 98.10 | 73.03 | 43.54 |
| C ``"`a <`"`` | 84.76 | 73.03 | 28.30 |
| D ``"`< b`"`` | 85.07 | 60.00 | 28.65 |
| F ``"`a < b < c`"`` | 126.42 | 73.03 | 75.91 |
| H `"a < b"` (plain) | 98.10 | 98.10 | 43.54 |

For F the box (73px) is now narrower than its label (76px) and the text visibly sticks out of the rect on both sides in Chromium; D gets a box sized for an empty label. Comparison operators in decision labels (`` {"`n < max?`"} ``) are common enough that this should not ship. The render side is fine (`html_text_content` only treats `<`+letter or `</`+letter as a tag, like `strip_html_for_measurement`); the layout side needs the same rule — e.g. make `concat_label_text_for_width` skip a `<` only when it is followed by an ASCII letter or `/`+letter, or have the stripper emit literal `<` as `&lt;` and decode entities in the width helper. Please add these to `flowchart_markdown_multiline_labels_fit_inside_their_box` (rect width must not shrink and must stay >= FO width + padding).

(`x < y > z` was already mis-measured on `main` because `strip_markdown_paragraph_for_measure` pairs the two brackets; that is pre-existing and not required here, but the same tag-start rule would fix it too.)

**Nits (non-blocking)**

- `is_blank_markup` uses `str::trim`, which strips U+00A0, so `a<br/>&nbsp;` counts 1 line while Chromium paints 2 (nbsp is a glyph). Unlikely in practice.
- `split_label_lines("<p>a</p>\n<p>b</p>")` returns 3 (the `\n` between blocks becomes a line). Not reachable today because `markdown_label_to_html` / `strip_markdown_for_measure` emit `</p><p>` without a newline, but worth a comment or a test pinning it.
- With a `font-size` style override the label group is still translated by the default-size half height (`translate(…, -24)` for a 72px FO at `font-size:24px`), so a 2-line label sits 12px low and pokes 1px out of the box. Pre-existing (`main` has the same `-12` for a 36px FO), just more visible now.

g65537 (changes requested):
Re-reviewed at `b08edf84` (base `4d4d7ad`). The last round's cases are fixed and the description's numbers hold, but the same mechanism still drops width for a markdown label whose `<` is followed by a letter, so one more pass is needed.

**What I verified**

- Built `main` (`4d4d7ad`; `0f3f2b64` renders identically) and this head, rendered all 1395 `.mmd` fixtures: 84 differ (39 cypress/flowchart, 30 demos/flowchart, 13 cypress/state, 2 demos/state), every one with a `<br>`, a label newline or a multi-line note. On those 84 the `foreignObject` width multiset and the `rect` width multiset are identical between `main` and this head; only heights move.
- Headless Chromium, declared `foreignObject` height vs painted div: files with a mismatch 408 -> 324, files with a label outside its shape 191 -> 121, exactly the 84 changed fixtures fixed, 0 worse, none of the 84 still mismatching.
- Last round's table is back to `main`: `` `a < b` `` 98.10, `` `a <` `` 84.76, `` `< b` `` 85.07, `` `a < b < c` `` 126.42, plain `a < b` 98.10; `` `x < y > z` `` 84.53 -> 125.18 (= the plain spelling). cypress/flowchart/126 (`2 < 4` in both spellings) keeps its `main` widths.
- `tag_len` agrees with Chromium's label div for `<3`, `a < b`, `x < y > z`, `<中> b`, `<<stereo>>` (paints `<>`), `<br class="x">`, `<BR>`, `</p>`; CJK widths are untouched (`中文<br/>标签` 20.34/77.80, `` `a < 中文` `` 132.49, same as `main`).
- Nits from last round: `a<br/>&nbsp;` now measures 2 lines (48px FO, 70.3 box; Chromium paints 2) and `<p>a</p>\n<p>b</p>` = 3 is pinned. Leaving the `font-size` override offset for its own change is fine.
- `cargo test -p mermaid-little`: 1041 pass. CI green.

**Blocking: markdown labels with `<` + letter still lose width (and lines)**

`tag_len` only refuses a `<` that is followed by a non-letter. `strip_markdown_for_measure` still wraps the paragraph in `<p>…</p>`, so when the source has `<` + letter and no `>` of its own (`i<n`, `count<max`, `x<y`), the wrapper's `</p>` again supplies the `>`, and `concat_label_text_for_width` skips from that `<` to the end of the paragraph. `main` measures these correctly because it has no wrapper, and the renderer (`markdown_label_to_html`) emits them as `&lt;`, so the browser paints the whole text:

```
flowchart TB
    A["`count<max`"]
    B["`if i<n then loop again`"]
    C["`x<y and y<z`"]
    D["`a<b

    second paragraph`"]
```

| node | rect width `main` | rect width `b08edf84` | FO width | Chromium |
|---|---|---|---|---|
| A `` `count<max` `` | 141.73 | 99.50 | 93.41 | fits by 6px |
| B `` `if i<n then loop again` `` | 211.68 | 77.16 | 173.34 | 48px outside the box |
| C `` `x<y and y<z` `` | 150.90 | 68.29 | 103.89 | 18px outside the box |
| D `` `a<b` + paragraph `` | 216.36 x 46.3 | 195.74 x 46.3 | 178.70 x 48 | 2 painted lines in a 1-line box |

D also shows the line count breaking: `split_label_lines("<p>a<b</p><p>second paragraph</p>")` takes `<b</p>` as one tag, misses the block end and returns 1. The plain-string spellings (`count<max` etc.) are unchanged from `main`. `` `i<n?` ``-style decision labels are at least as common as the spaced form, so this needs the same treatment.

The clean fix is to stop re-scanning literal text as markup: have `strip_markdown_paragraph_for_measure` emit a literal `<` as `&lt;` (exactly what `markdown_label_to_html` does for it) and decode entities on the markdown measurement path, or have the stripper return the plain text and the line count directly instead of a `<p>`-wrapped string. Please add `` `count<max` ``, `` `if i<n then loop again` `` and the two-paragraph `` `a<b` `` case to `flowchart_markdown_labels_measure_literal_less_than` (rect width >= FO width, and 2 lines for the paragraph case); the current cases all have a space after the `<`, which is why the test passes.

**Nits (non-blocking)**

- `A["a <!-- hidden --> b"]`: `string_label_to_html` passes `<!` through as a tag (Chromium paints `a  b`), while `tag_len` treats `<!` as text, so the box goes 86.37 (`main`) -> 192.57. Over-wide only, and HTML comments in labels are rare; treating `<!` as a tag in `tag_len` would align the two.
- `one<p>two` (string label) paints 2 lines in Chromium but `split_label_lines` counts 1: an opening `<p>` does not end the current block. Pre-existing.
- Pre-existing, not this PR: `` `x < y > z` `` is emitted raw (`x < y > z`) inside the `foreignObject`, which is not well-formed XML; a string label `count<max` is emitted as a raw `<max` tag and paints `count` only.

g65537 (changes requested):
Re-reviewed at `7e56834c` (base `4d4d7ad`). Round 3's blocker is genuinely gone and the description's numbers hold, but moving the FontAwesome stripping after the line concatenation introduces one regression against `main` that has to go first.

**What I verified**

- Built `main` (`4d4d7ad`) and this head, rendered all 1395 `.mmd` fixtures with both: 84 differ, the same 84 as last round (39 cypress/flowchart, 30 demos/flowchart, 13 cypress/state, 2 demos/state); `foreignObject` and `rect` width multisets are identical to `main` on every one of them.
- Headless Chromium, declared `foreignObject` height vs painted div: files with a mismatch 408 -> 324, files with a label outside its shape 191 -> 121, exactly the 84 changed fixtures fixed, 0 files worse, none of the 84 still mismatching.
- Round 3's table is back to `main`: `` `count<max` `` 141.73, `` `if i<n then loop again` `` 211.68, `` `x<y and y<z` `` 150.90, `` `a<b` `` + paragraph 216.36 x 70.3 with a 48px FO (2 lines). `` `i<n` `` measures the same in rect, diamond, hexagon, stadium, cylinder, circle and flag shapes as the plain spelling, and `` `**i<n**` ``, `` `<b>x</b> and i<n` ``, `` `a<br/>x<y` `` (48px FO, 70.3 box) all fit.
- Layout vs render agree on every markdown probe I tried: `<!-- c --> x`, `<3 you`, `x<y>z` (paints `xz`), `x < y > z`, `a<!b`, `a<!b> c`, entities (`&lt;`/`&amp;`/`&nbsp;` paint literally, measured literally), `**bold** *it*`, two paragraphs, `<br/>lead`, `trail<br/>`, `a<br/><br/>b`, raw `<strong>` + `<br>`, `a<p>b`, and CJK (`中文<br/>标签` 20.34 / 77.80, unchanged from `main`).
- `split_label_lines` rule changes checked in Chromium on the label div: `<p>a</p>\n<p>b</p>` paints 2 lines (the flipped test is right), `one<p>two` 2, `one<p>two<p>three` 3, `a <!-- hidden --> b` 1 line / `a  b`, `a<br/><!-- x -->` 1. `tag_len` treating `<!` as a tag matches `string_label_to_html`, and `A["a <!-- hidden --> b"]` now gets a 30px FO for a box of 86.37 (`main`: 151px FO in the same box).
- Plain string labels: identical to `main` on the sweep and on my probes except the intended `x < y > z` (84.53 -> 125.18) and the `<!` FO width above.
- `cargo test -p mermaid-little` 1042 pass; clippy `-D warnings` and fmt clean.

**Blocking: FontAwesome icon followed by a line break loses width (regression vs `main`)**

`measure_label_box` now runs `strip_fa_icons` on the *concatenated* plain text, after `concat_label_text_for_width` / `strip_markdown_for_measure` have removed the `<br/>` (or `\n`) that separated the icon name from the next line. The icon-name scan (`[a-z0-9-]+`) therefore runs straight into the following line's text and swallows its first word. `main` and `b08edf84` stripped icons on the raw markup first, where the `<` ends the name.

```
flowchart TB
    A["fa:fa-car<br/>Longer text here"]
    D["fa:fa-car<br/>x"]
    E["Car fa:fa-car<br/>Longer text here"]
    H["a<br/>fa:fa-car"]
    J["fa:fa-car<br/>fa:fa-bus"]
```

| node | rect width `main` | `b08edf84` | `7e56834c` | FO width | Chromium at `7e56834c` |
|---|---|---|---|---|---|
| A | 177.13 | 177.13 | 128.63 | 133.86 | text 2.6px outside the box |
| D | 68.29 | 68.29 | 60.00 | 9.47 | fits (min width) |
| E | 205.69 | 205.69 | 157.19 | 166.50 | text 4.7px outside the box |
| J | 60.00 | 60.00 | 108.33 | 0 | over-wide |

H is the same ordering problem on the line count: layout splits the raw `a<br/>fa:fa-car` into 2 lines (70.3px box) while the renderer's `split_label_lines` sees `<i class="fa fa-car"></i>` as blank markup and emits a 24px FO, so layout and render disagree here (both were 1 line at `b08edf84`).

An icon on its own line above the text is a common way to write these labels, and `main` sized them correctly, so this needs to go back to stripping the icon tokens before the lines are joined (per line, or on the raw label as `b08edf84` did) so that a `<br/>` / `\n` still terminates the icon name; then the line count and the width see the same text. A `fa:fa-car<br/>Longer text here` case in `diagram_bounds.rs` (rect width >= FO width + padding, 2 lines) would pin it. None of the 1395 fixtures contain `fa-name<br` (which is why the sweep is unaffected), so please probe it directly.

**Nits (non-blocking)**

- `` `fa:fa-car<br/>Longer text here` `` (markdown) already measured 128.63 on `main` for the same reason (`strip_markdown_paragraph_for_measure` drops the tag before the icon strip); `b08edf84` incidentally fixed it to 177.13. The per-line fix above covers it.
- `<p>a</p>\n\n<p>b</p>` counts 4 lines (Chromium: 2). Not reachable from any producer; a comment would do.

Everything else from rounds 1-3 holds at this head; this is the only remaining defect I found.

g65537 (approved):
Re-reviewed at `f5773eeb` (base `0f3f2b64`; `origin/main` `4d4d7ad` renders identically). Round 4's blocker is gone, the description's numbers hold, and I found no new defect, so approving.

**What I verified**

- Built base and head, rendered all 1395 `.mmd` fixtures with both: 84 differ, the same 84 as rounds 2-4 (39 cypress/flowchart, 30 demos/flowchart, 13 cypress/state, 2 demos/state); `foreignObject` and `rect` width multisets are identical to `main` on every one of them.
- Headless Chromium, declared `foreignObject` height vs painted div: files with a mismatch 408 -> 324, files with a label outside its shape 191 -> 121, exactly the 84 changed fixtures fixed, 0 files worse, none of the 84 still mismatching or overflowing.
- Round 4's table is back to `main`: `fa:fa-car<br/>Longer text here` 177.13 (FO 133.86 x 48, box 70.3), `fa:fa-car<br/>x` 68.29, `Car fa:fa-car<br/>...` 205.69, `fa:fa-car<br/>fa:fa-bus` 60 / FO 0; `a<br/>fa:fa-car` is one line on both sides (24px FO, 46.3 box). The markdown spelling `` `fa:fa-car<br/>Longer text here` `` goes 128.63 -> 177.13 (same as the plain spelling), and a backtick label with a real newline after the icon is also fixed.
- Icons in every position agree with Chromium and fit their box: alone on a line (leading, middle, trailing; `Start<br/>fa:fa-car<br/>End` = 3 lines like `a<br/><br/>b`), start/middle/end of a line, consecutive icons, icon + CJK, icon + `**bold**` in markdown, two-paragraph markdown, unknown icon names, `fa:` / `fa:fa-` with no name, uppercase `FA:FA-CAR`, `fa:fa-CAR`, trailing hyphen, `fab:`, and inside diamond / hexagon / stadium / cylinder / circle / flag / rounded shapes. Chromium's line boxes for the emitted `<i>` markup (`<i></i><br/>b` 2, `a<br/><i></i>` 1, `<br/><i></i>` 1, `<p><i></i></p><p>b</p>` 1, `<i></i><br/><br/>b` 3) match `split_label_lines`.
- Audited `measure_label_box` and its two callers: nothing scans across a break any more. `strip_fa_icons_per_line` splits on `\n` and `<br>` (via `tag_len` + `is_br_tag_body`) before stripping; `concat_label_text_for_width` only skips `tag_len` spans, which can no longer contain `<` or `\n`, and decodes nothing; `strip_markdown_paragraph_for_measure` works per paragraph and drops `\n` after `tag_len`; the line count comes from the emitted markup on both sides. `cargo test -p mermaid-little` 1044 pass, clippy `-D warnings` and fmt clean, CI green.

**Nits (non-blocking, all pre-existing behaviour or belong in their own change)**

- The two emitters still use their own tag rule: `string_label_to_html` scans a `<letter` tag to the first `>` regardless of what is in between, `markdown_paragraph_to_html` passes any `<...>` through. So `A["i<n<br/>loop"]` is still emitted as `i<n<br/>loop` (paints `iloop` on one line in an HTML context, as on `main`), while layout now counts two lines: box 93.8 x 46.3 / FO 38.6 x 24 on `main` -> 114.4 x 70.3 / FO 62.2 x 48. Over-sized only; the text loss is the pre-existing bug, and switching both emitters to `tag_len` (emit `i&lt;n<br/>loop`) would make the current layout exactly right.
- `tag_len` refusing a `<` inside a tag span affects an attribute value containing `<`: `A["<span title=\"a<b\">x</span>"]` measures 68.29 on `main`, 176.94 here (FO 133.6). Not well-formed XML in the first place; noting only.
- A string subgraph title with `<br/>` (`subgraph S ["Sub<br/>title"]`) is still painted as literal text on one line (known, same on `main`) but the cluster now reserves two: 96.3 -> 120.3 tall. The extra line will be needed once the title rendering is fixed, so it belongs with that change.
- `strip_fa_icons` strips a name-less `fa:fa-` that `replace_fa_icons` leaves as text (`fa:fa-<br/>Longer text here`: box 177.13 for a 175.9 FO), and the two use different name alphabets (`\w` vs `[A-Za-z0-9-]`, so `fa:fa-car_x` / `fa:fa-car中` measure wider than they paint). Identical to `main`.
- `<p>a</p>\n\n<p>b</p>` = 4 lines: agree it is unreachable; the comment is enough.

PR #244, opened 2026-09-23
chore: release font-metrics 0.1.1, mermaid-little 11.14.0-5, diagram 0.1.5

kookyleo:
Version bumps only — no code changes. This is what makes the two fixes merged since the last release reachable from crates.io.

| crate | from | to | why |
|---|---|---|---|
| `supramark-font-metrics` | 0.1.0 | 0.1.1 | #238 — missing East Asian Wide glyphs advance one em instead of the `.notdef` width (markon#102: overlapping CJK text in PlantUML) |
| `mermaid-little` | 11.14.0-4 | 11.14.0-5 | #239 — a label reserves one line per painted line (markon#97: multi-line node / edge / note labels spilling out of their shape) |
| `supramark-diagram` | 0.1.4 | 0.1.5 | carries the exact `mermaid-little` pin forward |

`plantuml-little` and `d2-little` depend on `supramark-font-metrics` through a caret range, so a downstream build resolves 0.1.1 without a release of their own. `supramark-diagram` is the only crate that pins `mermaid-little` exactly, hence its bump.

`cargo check -p supramark-diagram -p supramark-font-metrics -p mermaid-little --locked` passes; `Cargo.lock` updated in the same commit.

Publish order once merged: `supramark-font-metrics` → `mermaid-little` → `supramark-diagram`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #245, opened 2026-09-23
chore: release diagram-core 0.1.1 so mermaid-little 11.14.0-5 can build downstream

kookyleo:
Follow-up to #244, which I got wrong: `mermaid-little 11.14.0-5` and `supramark-diagram 0.1.5` are published but do not build for a downstream consumer.

## What is broken

`EngineRenderOptions` and `DiagramEngine::render_with_options` were added to `supramark-diagram-core` in #93, but that crate was never released — crates.io still carries 0.1.0, which has neither. Inside this workspace the path dependency hides it, and `cargo publish --no-verify` (my mistake) skipped the build that would have caught it.

Reproduced in a downstream crate (markon) after `cargo update`:

```
error[E0432]: unresolved import `supramark_diagram_core::EngineRenderOptions`
  --> mermaid-little-11.14.0-5/src/engine.rs
error[E0407]: method `render_with_options` is not a member of trait `DiagramEngine`
```

## Fix

| crate | from | to | why |
|---|---|---|---|
| `supramark-diagram-core` | 0.1.0 | 0.1.1 | ships the API #93 added |
| `mermaid-little` | 11.14.0-5 | 11.14.0-6 | requires `supramark-diagram-core = "0.1.1"` |
| `supramark-diagram` | 0.1.5 | 0.1.6 | same requirement, carries the exact mermaid pin forward |

Publishing 0.1.1 alone would be enough for a *fresh* resolution, since `^0.1` picks the newest compatible. It is not enough for an existing lock: `cargo update -p supramark-diagram` may keep `supramark-diagram-core` at 0.1.0, which still satisfies `^0.1`, and the build fails. Raising the requirement to `0.1.1` makes the resolver move it.

The diagram-core change is source compatible — the new trait method has a default body — so the already-published `plantuml-little 1.2026.2-4` and `d2-little 0.7.1-1` keep building against 0.1.1. Only the two crates that use the new API needed a requirement change.

`11.14.0-5` and `supramark-diagram 0.1.5` will be yanked once this is merged and the replacements are published.

`RUSTC_WRAPPER= cargo check --locked -p supramark-diagram -p mermaid-little -p supramark-diagram-core` passes; `Cargo.lock` updated in the same commit. This time every publish runs with verification enabled.


g65537 (approved):
Approving. I treated the description as a claim and re-derived each point locally; nothing blocking.

**Diagnosis confirmed.** `diff -r` of `~/.cargo/registry/src/.../supramark-diagram-core-0.1.0` against `crates/supramark-diagram-core` on this branch: the only non-rustfmt differences are the additions of `pub struct EngineRenderOptions`, `DiagramEngine::render_with_options` (with a default body that delegates to `render`), and `DiagramRegistry::render_with_options`. No existing public item changed shape, so 0.1.1 is the right bump and an external `DiagramEngine` implementor written against 0.1.0 still compiles.

**Reproduced the breakage as published.** A scratch crate depending on `supramark-diagram = "=0.1.5"` from crates.io (no patches) fails exactly as described:
```
error[E0432]: unresolved import `supramark_diagram_core::EngineRenderOptions`  --> mermaid-little-11.14.0-5/src/engine.rs:14
error[E0407]: method `render_with_options` is not a member of trait `DiagramEngine` --> mermaid-little-11.14.0-5/src/engine.rs:55
```
so yanking `mermaid-little 11.14.0-5` and `supramark-diagram 0.1.5` after the replacements are up is the right call.

**Verified the post-publish combination empirically.** `cargo package --no-verify` for the three crates on this branch, extracted the `.crate` files (their rewritten manifests carry `supramark-diagram-core = "0.1.1"` and `mermaid-little = "=11.14.0-6"`), and pointed a scratch consumer at them via `[patch.crates-io]` with *everything else from crates.io* (`plantuml-little 1.2026.2-4`, `d2-little 0.7.1-1`, `graphviz-anywhere 0.2.6`, `supramark-font-metrics 0.1.1`, `dagre 0.1.1`). It builds, and `default_registry()` renders mermaid via `render_with_options`, plantuml and d2 successfully. That is the claim that the already-published plantuml/d2 keep building against diagram-core 0.1.1, checked by compiling rather than by reading the trait.

**Stale-lock case.** Took the lock from the failing consumer above (diagram 0.1.5 / mermaid 11.14.0-5 / core 0.1.0), switched the requirement to `0.1.5`, ran `cargo update -p supramark-diagram`: cargo removes all three old entries and locks 0.1.6 / 11.14.0-6 / 0.1.1, then builds. The `^0.1.1` requirement (not just publishing 0.1.1) is what forces that.

**Release mechanics.** `RUSTC_WRAPPER= cargo metadata --locked` passes. crates.io has core 0.1.0 only, mermaid up to 11.14.0-5, diagram up to 0.1.5, so 0.1.1 / 11.14.0-6 / 0.1.6 are free. `cargo publish --dry-run -p supramark-diagram-core` **with verification** packages 5 files and builds. The dry-runs for mermaid-little and supramark-diagram cannot run yet: they fail at resolution (`supramark-diagram-core = "^0.1.1"` / `mermaid-little = "=11.14.0-6"` not on crates.io), which is expected and is exactly why the publish order must be core -> mermaid -> diagram; the patched consumer build above is the stand-in for those two dry-runs. `cargo package --list` for all three looks sane (core: 5 files; mermaid: 249; diagram: 7).

**Is this the whole problem?** I compared every published crate's registry source against the workspace tree:
- identical to tree: `graphviz-anywhere 0.2.6`, `supramark-font-metrics 0.1.1`, `dagre 0.1.1`, `mermaid-little 11.14.0-5` (src identical; 11.14.0-6 is a manifest-only republish), `supramark-diagram 0.1.5`.
- tree has unreleased changes: `plantuml-little 1.2026.2-4` (activity/sequence layout, parser), `d2-little 0.7.1-1` (lib.rs, semantic.rs, dagre_layout.rs), `supramark-markdown 0.1.2`. None of that is API another to-be-published crate depends on: `render_with_options`/`EngineRenderOptions` are used only by mermaid-little and supramark-diagram (grep), supramark-diagram touches plantuml/d2 only through `PlantumlEngine` / `D2Engine`, and the patched build above compiles it against the *published* plantuml/d2. `font-metrics 0.1.0 -> 0.1.1` adds no public API (private `is_east_asian_wide` plus behaviour), and `graphviz-anywhere 0.2.5 -> 0.2.6` changes only `build.rs`, so the `^0.1` / `^0.2` requirements elsewhere cannot mask a compile-time gap. No second instance of the diagram-core problem exists.

Merging with a merge commit after the two remaining `Test & Build` jobs finish, as with #237/#238/#239/#244.

Non-blocking notes in a separate comment.

g65537:
Non-blocking notes from the review (none of these should hold up the release):

1. **`EngineRenderOptions` is not `#[non_exhaustive]`.** It has a public field and is constructed by struct literal (`EngineRenderOptions { edge_label_decluster: true }` in mermaid-little's tests). Once 0.1.1 is out, adding a second option is a breaking change for any downstream that does the same. Marking it `#[non_exhaustive]` (and constructing via `..Default::default()`) would cost nothing now and a semver-major later. Fine to do in 0.1.2 before anyone depends on the literal, but it is cheapest before 0.1.1 ships.

2. **plantuml-little and d2-little carry unreleased fixes.** The registry sources of `plantuml-little 1.2026.2-4` and `d2-little 0.7.1-1` differ from the tree (plantuml: activity/sequence layout, parser, renderer — including the #238 sequence text-overlap fix; d2: `lib.rs`, `semantic.rs`, `dagre_layout.rs`). They do not need anything unreleased and the patched consumer build compiles supramark-diagram 0.1.6 against the published ones, so this is not a blocker. But the `font-metrics 0.1.1` CJK change released in #244 only takes effect for plantuml's `Notdef` policy, and downstream users still get the old plantuml. A `plantuml-little 1.2026.2-5` (and a `d2-little` bump) would close that. `supramark-markdown 0.1.2` is in the same state, though nothing in this release path depends on it.

3. **`supramark-diagram-core` manifest has no `repository` / `documentation` / `homepage`.** `cargo publish --dry-run` warns about it; worth adding before 0.1.1 or in the next release so the crates.io page links back here.

4. **CI gate for this class of defect.** The workspace `path` dependencies mean `cargo check` can never see that a sibling's *published* version lacks an API. The check that would have caught #244 is the one I ran by hand: build each to-be-published crate as it will exist on crates.io. Concretely, a `release-check` job that, for every crate in `crates/*` with `publish != false`:
   ```
   cargo package -p <crate> --no-verify        # rewrites path deps to registry requirements
   ```
   extracts `target/package/<crate>-<ver>.crate` into a temp dir, and then `cargo check`s a scratch consumer whose `[patch.crates-io]` points at *all* the extracted packages (so sibling crates bumped in the same PR resolve), while every other dependency comes from crates.io. If a crate's rewritten requirement selects a published version that lacks an API the crate uses, that step fails with the exact E0432/E0407 seen downstream. Cheaper but narrower alternative: a script that, for every `foo = { path = "...", version = "req" }`, asserts either that the sibling crate's tree equals the registry source of the newest published version matching `req`, or that `req` admits no published version (i.e. the sibling is being bumped in the same PR). And, independently, `cargo publish` in the release workflow should never pass `--no-verify`.

PR #247, opened 2026-09-23
chore(diagram-core): seal EngineRenderOptions before it is published

kookyleo:
Takes nits 1 and 3 from the #245 review. Both are things a publish cannot undo, and neither is published yet — crates.io still has `supramark-diagram-core 0.1.0`, which predates `EngineRenderOptions` entirely — so this is the last cheap moment for either.

## `#[non_exhaustive]` on `EngineRenderOptions`

The struct is a bag of optional readability flags, so a second flag is a question of when, not whether. Without `#[non_exhaustive]`, adding one breaks every struct literal in every downstream crate and forces a major release of diagram-core.

Sealing it costs one thing: a literal is no longer available outside the crate, and `..Default::default()` is not a way around that for a `#[non_exhaustive]` struct. So this adds `with_edge_label_decluster`, and the single external literal (a mermaid-little test) now reads:

```rust
let opts = EngineRenderOptions::default().with_edge_label_decluster(true);
```

`EngineRenderOptions::default()` was already the spelling everywhere else.

## Manifest metadata

`supramark-diagram-core` was missing `homepage` and `repository` — `cargo publish --dry-run` warns about it, and a published version cannot gain them. Same values as `mermaid-little` and `supramark-font-metrics`.

## No version changes

0.1.1 and 11.14.0-6 are still unpublished, so this edits them in place rather than bumping again.

`cargo clippy -p supramark-diagram-core -p mermaid-little --all-targets -- -D warnings` and `cargo fmt --check` are clean.

Unrelated note: a workspace-wide clippy on this machine also reports 8 pre-existing errors in `plantuml-little` (unused doc comment, collapsible `if`, `map_or`, four `?`-operator blocks, a needless `&mut` on `ensure_create_participant`), all in tree changes that predate this branch and none of them in a crate this PR touches. CI is green, so it is presumably a newer local clippy; noting it so it is not a surprise later.


g65537 (approved):
Approving. I treated each claim as something to re-derive locally rather than accept; nothing blocking.

**Premise holds.** crates.io API (`/api/v1/crates/<name>/versions`) today: `supramark-diagram-core` has only `0.1.0`; `mermaid-little` tops out at `11.14.0-5`; `supramark-diagram` at `0.1.5`. So `0.1.1` / `11.14.0-6` / `0.1.6` are unpublished and editing them in place without a bump is correct. I also pulled the three published `.crate` files and grepped them: `diagram-core 0.1.0` has no `EngineRenderOptions` at all, and the only struct literal of it anywhere on crates.io is the one in `mermaid-little 11.14.0-5`'s `#[cfg(all(test, feature = "metrics-ttf-parser"))]` module, which no consumer compiles. Its non-test code only reads `opts.edge_label_decluster`, which `#[non_exhaustive]` does not affect. So sealing the struct now breaks nobody, published or not.

**Every construction site still compiles.** `grep -rn EngineRenderOptions` over the workspace finds exactly: the trait/registry signatures and setter in diagram-core, `EngineRenderOptions::default()` and the new `default().with_edge_label_decluster(true)` in mermaid-little's tests, and a bare re-export in supramark-diagram. No exhaustive `match`/destructure anywhere. `cargo clippy -p supramark-diagram-core -p mermaid-little --all-targets -- -D warnings` and `cargo fmt --check` clean; `cargo test -p supramark-diagram-core` 3/3; `cargo test -p mermaid-little --lib render_options_tests` 2/2 both with default features and with `--no-default-features` (the dev-dep self-cycle still turns on `metrics-ttf-parser`, so the test is not silently skipped). The rewritten test still asserts what it did (SVG mime + `<svg` with decluster on).

**Escape hatch is sufficient, checked from outside the workspace.** A scratch crate under `/tmp` with a `path` dependency on the branch's diagram-core: `EngineRenderOptions::default().with_edge_label_decluster(true)`, direct field assignment through a `mut` binding, and the `{ edge_label_decluster, .. }` rest pattern all compile and run; a struct literal, `..Default::default()` functional update, and an exhaustive `{ edge_label_decluster }` pattern are rejected with E0639 / E0639 / E0638 respectively. With one `bool` field, `Default` plus the setter (or field assignment) reaches every value of the type. The `with_*` by-value `mut self -> Self` shape matches `RenderOptions::with_edge_label_decluster` in mermaid-little and the `with_*` setters in font-metrics / plantuml-little.

**Manifest.** `cargo publish -p supramark-diagram-core --dry-run` (with verification) packages 5 files and no longer warns; `RUSTDOCFLAGS=-D warnings cargo doc -p supramark-diagram-core` is clean, so the `Self::edge_label_decluster` intra-doc link resolves. `homepage`/`repository` values match mermaid-little / plantuml-little / d2-little.

**Pre-existing clippy noise confirmed.** `git diff origin/main..HEAD` touches only the three files in this PR; a workspace-wide `cargo clippy --all-targets -- -D warnings` on local clippy 0.1.97 fails in `dagre` (2 errors: `sort_by_key`, manual `Option::filter`) and in `plantuml-little` (the 8 you listed); `cargo clippy -p plantuml-little` on `main` reproduces the same 8. Neither crate is touched here, and CI's pinned clippy is green.

Merging with a merge commit once the two `Test & Build` jobs finish, as with #237/#238/#239/#244/#245. Non-blocking notes in a separate comment.

g65537:
Non-blocking notes. The first one is the only "now or never" item I see; the rest can be taken in any later patch release.

1. **`Copy` and `Eq` on `EngineRenderOptions` are also one-way doors.** `#[non_exhaustive]` makes adding a field non-breaking, but only if the new field keeps every derived trait: a `String`/`Vec` field forces dropping `Copy`, an `f32`/`f64` (say a label-padding factor) forces dropping `Eq`, and removing either derive is itself a breaking change. If you are confident every future hint is a `bool`/small `Copy` scalar, keep them; if not, dropping `Copy` (and possibly `Eq`) before 0.1.1 is free today and a semver-major later. The trait takes `&EngineRenderOptions`, so nothing in the workspace needs `Copy`. Adding a derive back later is non-breaking, so erring on the side of fewer derives costs nothing.

2. `with_edge_label_decluster` could be `pub const fn`, matching mermaid-little's own `RenderOptions::with_edge_label_decluster` (which already compiles on the 1.82 MSRV). Adding `const` later is non-breaking, so purely cosmetic.

3. Manifest: still no `documentation = "https://docs.rs/supramark-diagram-core"` (supramark-diagram has it), `keywords`, `categories`, or `readme`. None is irreversible: they can land with the next version.

4. Reminder for the actual publish, since this PR now makes the three tree versions final: order must be `supramark-diagram-core 0.1.1` -> `mermaid-little 11.14.0-6` -> `supramark-diagram 0.1.6`, then yank `mermaid-little 11.14.0-5` and `supramark-diagram 0.1.5` (both import `EngineRenderOptions` from a `^0.1` core that only has 0.1.0 on crates.io, so they cannot build today; after 0.1.1 ships they resolve to it and build again, which is fine but the yank keeps `cargo update` from ever selecting a pair that was never tested together).

